PR #2, opened 2015-01-03
Update linux-bootstrap-1.md

Streppel:
Some minor english enhancements. Didn't go much further on the file because I was not sure if I'd change some content unintentionally


linux-bootstrap-1.md
```diff
@@ -386,7 +386,7 @@ Let's look on all of these cases:
 	sti
 ```
 
-Here we can see aligning of `dx` (contains `sp` given by bootloader) to 4 bytes and checking that it is not zero. If it is zero we put `0xfffc` (4 byte aligned address before maximum segment size - 64 Kb) to `dx`. If it is not zero we continue to use `sp` given by bootloader (0xf7f4 in my case). After this we put `ax` value to `ss` which stores correct segment address `0x10000` and set up correct `sp`. After it we have correct stack:
+Here we can see aligning of `dx` (contains `sp` given by bootloader) to 4 bytes and checking that it is not zero. If it is zero we put `0xfffc` (4 byte aligned address before maximum segment size - 64 Kb) to `dx`. If it is not zerowe  continue to use `sp` given by bootloader (0xf7f4 in my case). After this we put `ax` value to `ss` which stores correct segment address `0x10000` and set up correct `sp`. After it we have correct stack:
```
0xAX, line 389: zerowe?

Seems that ',' must be there

Streppel:
Oh, zerowe was a mistake on my part. My apologies.

0xAX:
@Streppel I close this PR to not mix all PRs, please feel free to reopen it if there will be updates

PR #9, opened 2015-01-04
Changed magic-byte and swapped exponent

ckrooss:


0xAX:
ah, you're about it.

sorry, content is correct. Intel's CPUs uses little-endian notation, it means that low byte will be first in order, for example `0xaa55` will be `0x55aa` in little endian. 

that would not be confused you can just define it:

```
dw 0xaa55
```

in your code.

You can read more [here](http://en.wikipedia.org/wiki/Endianness).

ckrooss:
Endianness, my old nemesis, we meet again.

0xAX:
Thank you @ckrooss merged

PR #10, opened 2015-01-04
Improve English throughout the file

CaptainChloride:
Made as many edits as I could find. Hope it helped! Good job with this, it really helped. 

Thanks :-)

### DIFF
--- a/linux-bootstrap-1.md
+++ b/linux-bootstrap-1.md
@@ -4,7 +4,7 @@ GNU/Linux kernel internals
 Linux kernel booting process. Part 1.
 --------------------------------------------------------------------------------
 
-If you have read my previous [blog posts](http://0xax.blogspot.com/search/label/asm), you can see that some time ago I started to get involved with low-level programming. I wrote some posts about x86_64 assembly programming for Linux. At the same time, I started to dive into the GNU/Linux kernel source code. It is very interesting for me to understand how low-level things work, how programs run on my computer, how they are located in memory, how the kernel manages processes and memory, how the network stack works on low-level and many many other things. I decided to write yet another series of posts about the GNU/Linux kernel for **x86_64**.
+If you read my previous [blog posts](http://0xax.blogspot.com/search/label/asm), you'll see that some time ago I started to get involved with low-level programming. I wrote some posts about x86_64 assembly programming for Linux. At the same time, I started to dive into the GNU/Linux kernel source code. It is very interesting for me to understand how low-level things work, how programs run on my computer, how they are located in memory, how the kernel manages processes and memory, how the network stack works on low-level and many many other things. I decided to write yet another series of posts about the GNU/Linux kernel for the **x86_64** processors.
 
 Note that I'm not a professional kernel hacker, and I don't write code for the kernel at work. It's just a hobby. I just like low-level stuff, and it is interesting for me to see how these things work. So if you notice anything confusing, or if you have any questions/remarks, ping me on twitter [0xAX](https://twitter.com/0xAX), drop me an [email](anotherworldofworld@gmail.com) or just create an [issue](https://github.com/0xAX/linux-internals/issues/new). I appreciate it. All posts will also be accessible at[linux-internals](https://github.com/0xAX/linux-internals) and if you find something wrong with my English or post content, feel free to send pull request.
 
@@ -16,17 +16,16 @@ Note that I'm not a professional kernel hacker, and I don't write code for the k
 * Understanding C code
 * Understanding assembly code (AT&T syntax)
 
-Anyway, if you just started to learn some tools, I will try to explain some parts during this and following posts. Ok, little introduction finished and now we can start to dive into kernel and low-level stuff.
+Anyway if you just started to learn some tools, I will try to explain some parts during this and the following posts. Ok, enough with the introductions. Lets dive into kernel and low-level stuff.
 
-All code is actual for kernel - 3.18, if there will be changes, I will update posts.
+All code is for kernel - 3.18, if there come about any changes, I will update the posts.
 
 Magic power button, what's next?
 --------------------------------------------------------------------------------
 
-Despite that it is series of posts about linux kernel, we will not start from kernel code (at least in this paragraph). Ok, you pressed magic power button on your laptop or desktop computer and it started to work. After this mother board sends signal to the [power supply](http://en.wikipedia.org/wiki/Power_supply) which provides computer with the proper amount of electricity. Once motherboard receives [power good signal](http://en.wikipedia.org/wiki/Power_good_signal), it tries to run CPU. CPU resets all leftover data in its registers and sets up predefined values for every register.
+Despite that it is a series of posts about the Linux kernel, we will not start with kernel code (at least in this paragraph). Ok, so you pressed the magic power button on your laptop or desktop computer and it started up. After the motherboard sends signal to the [power supply](http://en.wikipedia.org/wiki/Power_supply), it provides the computer with the proper amount of electricity. Once the motherboard receives the [power good signal](http://en.wikipedia.org/wiki/Power_good_signal), it tries to run the CPU. The CPU then resets all of the leftover data in its register and sets up predefined values for every register.
 
-
-[80386](http://en.wikipedia.org/wiki/Intel_80386) and later CPUs defines following predifined data in CPU registers after computer resets:
+[80386](http://en.wikipedia.org/wiki/Intel_80386) and later CPUs define the following predifined data in CPU registers after computer resets:
 
 ```
 IP          0xfff0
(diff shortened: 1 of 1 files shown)


linux-bootstrap-1.md
```diff
@@ -446,17 +444,17 @@ Ok now we have correct segment registers, stack, need only setup bss and jump to
...
 
 Jump to main
 --------------------------------------------------------------------------------
 
-That's all, we have stack, bss and now we can jump to `main` C function:
+That's all! We have our stack and BSS set up. Now we can jump to the `main` C function:
 
 ```assembly
-	calll main
+	call main
```
0xAX, line 457: It's not error, it's correct `gas` instruction

linux-bootstrap-1.md
```diff
@@ -165,24 +163,24 @@ General real mode memory map is:
 0x000F0000 - 0x000FFFFF - System BIOS
 ```
 
-But stop, at the beginning of post I wrote that first instruction executed by CPU located by `0xfffffff0` address, which is much bigger than `0xffff` (1MB). How can CPU access it in real mode? As I write about and you can read in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
+But stop; at the beginning of this post, I had written that the first instruction executed by the CPU is located by `0xfffffff0` address; however, it's much bigger than `0xffff` (1MB). How can the CPU access *this* in real mode? As I write and you can read about in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
```
0xAX, line 166: Why `;`? Why not `.` or `,`?

linux-bootstrap-1.md
```diff
@@ -130,26 +128,26 @@ We will see:
...
-as I wrote above. But we have only 16 bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
+As i wrote above. But we have only 16-bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
 
 ```python
 >>> hex((0xffff * 16) + 0xffff)
 '0x10ffef'
 ```
 
-Where `0x10ffef` is equal to `1mb + 64KB - 16b`. But [8086](http://en.wikipedia.org/wiki/Intel_8086) processor, which was first processor with real mode, had 20 bit address line, and `2^20 = 1048576.0` which is 1MB, so it means that actually available memory amount is 1MB.
+Where `0x10ffef` is equal to `1mb + 64KB - 16b`. But the [8086](http://en.wikipedia.org/wiki/Intel_8086) processor, which was first processor with real mode had 20 address lines, but `20^2 = 1048576.0` which is 1MB; this means that the amount of memory actually available was 1MB.
```
0xAX, line 148: This paragraph was edited already.

linux-bootstrap-1.md
```diff
@@ -130,26 +128,26 @@ We will see:
...
 
-**NOTE**: as you can read above CPU is in real mode. In real mode for calculating physical address in memory uses following form:
+**NOTE**: as you can read above, the CPU is in real mode. In real mode, for calculating physical address of memory, it uses the following form:
 
 ```
 PhysicalAddress = Segment * 16 + Offset
 ```
 
-as I wrote above. But we have only 16 bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
+As i wrote above. But we have only 16-bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
```
0xAX, line 141: Somebody changes i -> I, somebody I -> i, i don't know how to be...

linux-bootstrap-1.md
```diff
@@ -130,26 +128,26 @@ We will see:
...
+In this example we can see that this code will be executed in 16-bit real mode and started at 0x7c00 in memory. After the start it calls [0x10](http://www.ctyme.com/intr/rb-0106.htm) interruption which just prints `!` symbol. It fills rest of the 510 bytes with zeros and ends it with two magic bytes 0xaa and 0x55.
 
-Real world boot loader starts at the same point, ends with `0xaa55` bytes, but reads kernel code from device, loads it to memory, parses and passes boot parameters to kernel and etc... intead of printing one symbol :) Ok, so, from this moment bios handed control to the operating system bootloader and we can go ahead.
+In the real world, the bootloader starts at the same point, ends with `0xaa55` bytes, but reads kernel code from the device instead, loading it to memory, parsing it and passing the necessary boot parameters to the kernel, among other required tasks... intead printing one symbol :) Ok, so, from this moment, the BIOS has officially handed control to the operating system bootloader and we can go ahead.
```
0xAX, line 133: intead -> instead

also see previous pull requets for this paragraph.

linux-bootstrap-1.md
```diff
@@ -106,21 +105,20 @@ Now BIOS started to work, after all initializations, hardware checking, it needs
...
 
-    int 0x10
+	int 0x10
     jmp $
 
 times 510-($-$$) db 0
-
-db 0x55
 db 0xaa
+db 0x55
```
0xAX, line 118: Must be:

```
db 0x55
db 0xaa
```

see previous commit

linux-bootstrap-1.md
```diff
@@ -106,21 +105,20 @@ Now BIOS started to work, after all initializations, hardware checking, it needs
...
 
 boot:
-    mov al, '!'
     mov ah, 0x0e
     mov bh, 0x00
     mov bl, 0x07
+    mov al, !
 
-    int 0x10
+	int 0x10
```
0xAX, line 113: yes, tabs are better for assembly code, but they must be everywhere

linux-bootstrap-1.md
```diff
@@ -83,7 +82,7 @@ reset_vector:
 	...
 ```
 
-We can see here jump instruction [opcode](http://ref.x86asm.net/coder32.html#xE9) - 0xe9 to the address `_start - ( . + 2)`. And we can see that `reset` section is 16 bytes and starts at `0xfffffff0`:
+We can see here that the jump instruction [opcode](http://ref.x86asm.net/coder32.html#xE9) - 0xe9 to the address `_start - ( . + 2)`. And we can see that `reset` section is 16 bytes and starts at `0xfffffff0`:
```
0xAX, line 85: I think no need in `that` here

linux-bootstrap-1.md
```diff
@@ -54,24 +53,24 @@ But if we take the biggest segment part and offset: `0xffff:0xffff`, it will be:
...
-which we can translate to the physical address::
+which we can translate to the physical address:
 
 ```python
 >>> hex((0xffff000 << 4) + 0xfff0)
 '0xfffffff0'
 ```
 
-We get `fffffff0` which is 4GB - 16 bytes. This point is the [Reset vector](http://en.wikipedia.org/wiki/Reset_vector). This is the memory location at which CPU expects to find the first instruction to execute after reset. It contains [jump](http://en.wikipedia.org/wiki/JMP_%28x86_instruction%29) instruction which usually points to the BIOS entry point. For example if we look in [coreboot](http://www.coreboot.org/) source code, we will see it:
+We get `fffffff0` which is 4GB - 16 bytes. This point is a - [Reset vector](http://en.wikipedia.org/wiki/Reset_vector). TThe first instruction exists at this memory location, which the CPU inteprets after reset. It contains the [jump](http://en.wikipedia.org/wiki/JMP_%28x86_instruction%29) instruction which usually points to the BIOS entry point. For example if we look into the [coreboot](http://www.coreboot.org/) source code, we see:
```
0xAX, line 73: CPU doesn't interprets, it executes

linux-bootstrap-1.md
```diff
@@ -54,24 +53,24 @@ But if we take the biggest segment part and offset: `0xffff:0xffff`, it will be:
 '0x10ffef'
 ```
 
-which is 65519 bytes over first megabyte. Since only one megabyte is accessible in real mode, `0x10ffef` becomes `0x00ffef` with disabled [A20](http://en.wikipedia.org/wiki/A20_line).
+which is 65519 bytes over the first megabyte. Since only one megabyte accessible in real mode, `0x10ffef` becomes `0x00ffef` with disabled [A20](http://en.wikipedia.org/wiki/A20_line).
 
-Ok, now we know about real mode and memory addressing, let's get back to register values after reset.
+Ok, now we know about real mode and memory addressing, let's back to registers values after reset.
```
0xAX, line 58: the same https://github.com/0xAX/linux-insides/pull/7/files#diff-054c5daca2ffd0c4037409522c210d3cR59

linux-bootstrap-1.md
```diff
@@ -34,13 +33,13 @@ CS selector 0xf000
...
 
 ```
 PhysicalAddress = Segment * 16 + Offset
 ```
 
-For example `CS:IP` is `0x2000:0x0010`, physical address will be:
+For example `CS:IP` is `0x2000:0x0010`; its physical address will be:
```
0xAX, line 42: Why `;`

linux-bootstrap-1.md
```diff
@@ -16,17 +16,16 @@ Note that I'm not a professional kernel hacker, and I don't write code for the k
 * Understanding C code
 * Understanding assembly code (AT&T syntax)
 
-Anyway, if you just started to learn some tools, I will try to explain some parts during this and following posts. Ok, little introduction finished and now we can start to dive into kernel and low-level stuff.
+Anyway if you just started to learn some tools, I will try to explain some parts during this and the following posts. Ok, enough with the introductions. Lets dive into kernel and low-level stuff.
```
0xAX, line 19: return `,` plese

linux-bootstrap-1.md
```diff
@@ -4,7 +4,7 @@ GNU/Linux kernel internals
...
+If you read my previous [blog posts](http://0xax.blogspot.com/search/label/asm), you'll see that some time ago I started to get involved with low-level programming. I wrote some posts about x86_64 assembly programming for Linux. At the same time, I started to dive into the GNU/Linux kernel source code. It is very interesting for me to understand how low-level things work, how programs run on my computer, how they are located in memory, how the kernel manages processes and memory, how the network stack works on low-level and many many other things. I decided to write yet another series of posts about the GNU/Linux kernel for the **x86_64** processors.
```
0xAX, line 7: @ThomasMarcelis already fixed it, see https://github.com/0xAX/linux-insides/pull/1/files, I'm not english man and don't know what version is correct.

0xAX:
Hello @CaptainChloride, thank you for you work.

Please, see, i left some comments

PR #36, opened 2015-01-18
Just one typo fixed and one sentece time corrected

alx741:
- A typo in a hex number fixed
- Guessing that a sentece at the foot in -Conclusions- section should be better expressed in past time


0xAX:
Thank you @alx741, merged

PR #41, opened 2015-01-20
Some misspellings corrected through the part 2 document

alx741:


Huddayn:
```
 - Memory segmentation was completely redone in the protected mode.
 + Memory segmentation was completely remade in the protected mode.
```

I find 'redone' more fitting here.

alx741:
Thank you @Huddayn , changing...

alx741:
So 'its', 'is', and 'an' are the changes, someone can confirm if its better?

0xAX:
@alx741 @Haddayn Thank you.

@alx741 I like this changes, can you please rebase on master and i will merge it.

linux-bootstrap-2.md
```diff
@@ -92,7 +92,7 @@ Don't worry, i know that it looks little scary after real mode, but it's easy. L
 | 15          1    1    1   1 | Code            | Execute/Read, conforming, accessed
 ```
 
-As we can see first bit is 0 for data segment and 1 for code segment. Next three bits `EWA` are expansion direction (expand-down segment will grow down, more about it you can read [here](http://www.sudleyplace.com/dpmione/expanddown.html)), write enable and accessed for data segments. `CRA` bits are conforming (A transfer of execution into a more-privileged conforming segment allows execution to continue at the current privilege level), read enable and accessed.
+As we can see first bit is 0 for data segment and 1 for code segment. Next three bits `EWA` are expansion direction (expand-dow segment will grow down, you can read more about this [here](http://www.sudleyplace.com/dpmione/expanddown.html)), write enable and accessed for data segments. `CRA` bits are conforming (A transfer of execution into a more-privileged conforming segment allows execution to continue at the current privilege level), read enable and accessed.
```
0xAX, line 95: There is must be expand-down

alx741:
hmm... I've actually introduced new misspellings... closing and re-trying latter with new stuff

PR #43, opened 2015-01-21
Better description to how cpu starts

leandromoreira:


0xAX:
Thank you @leandromoreira, merged

PR #44, opened 2015-01-21
Fixes call opcode

leandromoreira:


0xAX:
@leandromoreira it's not error.

PR #49, opened 2015-02-02
Make links to other parts relative

Huddayn:
Links to other parts had absolute URLs, so they were pointing to the
same page, regardless of where they were located. E.g. pages on gitbooks
links to github.
Make them relative, so they are pointing to the host on which the are
located.


0xAX:
Thank you @Huddayn

PR #51, opened 2015-02-02
Booting grammar fixes

georgehorrell:
Some minor changes to the grammar in linux-bootstrap-1, hopefully improves the readability a little.

Cheers,
George

### DIFF
--- a/Booting/linux-bootstrap-1.md
+++ b/Booting/linux-bootstrap-1.md
@@ -1,29 +1,31 @@
-Kernel booting process. Part 1.
+Kernel Booting Process. Part 1.
 ================================================================================
 
-From the bootloader to kernel
+From the Bootloader to the Kernel
 --------------------------------------------------------------------------------
 
-If you have read my previous [blog posts](http://0xax.blogspot.com/search/label/asm), you can see that some time ago I started to get involved with low-level programming. I wrote some posts about x86_64 assembly programming for Linux. At the same time, I started to dive into the Linux source code. It is very interesting for me to understand how low-level things work, how programs run on my computer, how they are located in memory, how the kernel manages processes and memory, how the network stack works on low-level and many many other things. I decided to write yet another series of posts about the Linux kernel for **x86_64**.
+If you have read my previous [blog posts](http://0xax.blogspot.com/search/label/asm), you can see that some time ago I started to get involved with low-level programming. I wrote some posts about x86_64 assembly programming for Linux. At the same time, I started to dive into the Linux source code. It is very interesting for me to understand how low-level things work, how programs run on my computer, how they are located in memory, how the kernel manages processes and memory, how the network stack works on low-level and many many other things. I have therefore decided to write another series of posts about the Linux kernel for **x86_64**.
 
-Note that I'm not a professional kernel hacker, and I don't write code for the kernel at work. It's just a hobby. I just like low-level stuff, and it is interesting for me to see how these things work. So if you notice anything confusing, or if you have any questions/remarks, ping me on twitter [0xAX](https://twitter.com/0xAX), drop me an [email](anotherworldofworld@gmail.com) or just create an [issue](https://github.com/0xAX/linux-insides/issues/new). I appreciate it. All posts will also be accessible at [linux-insides](https://github.com/0xAX/linux-insides) and if you find something wrong with my English or post content, feel free to send pull request.
+Please note that I'm not a professional kernel hacker, and I don't write code for the kernel at work. It's just a hobby; I like low-level stuff, and it is interesting for me to see how these things work. If you notice anything confusing, or if you have any questions/remarks, ping me on twitter [0xAX](https://twitter.com/0xAX), drop me an [email](anotherworldofworld@gmail.com) or just create an [issue](https://github.com/0xAX/linux-insides/issues/new). I would really appreciate it. All posts will also be accessible at [linux-insides](https://github.com/0xAX/linux-insides) and if you find something wrong with my English or post content, feel free to send pull request.
 
 
 *Note that this isn't official documentation, just learning and sharing knowledge.*
 
 **Required knowledge**
 
-* Understanding C code
-* Understanding assembly code (AT&T syntax)
+* An understanding of C code
+* An understanding of assembly code (AT&T syntax)
 
-Anyway, if you just started to learn some tools, I will try to explain some parts during this and following posts. Ok, little introduction finished and now we can start to dive into kernel and low-level stuff.
+If you just started to learn some tools, I will try to explain some parts during this and the following posts. Ok, our little introduction is finished and now we can start to dive into th kernel and low-level stuff.
 
-All code is actual for kernel - 3.18, if there are changes, I will update posts.
+All code is actually for kernel - 3.18, if there are changes, I will update my posts.
 
 Magic power button, what's next?
 --------------------------------------------------------------------------------
 
-Despite that this is a series of posts about linux kernel, we will not start from kernel code (at least in this paragraph). Ok, you pressed magic power button on your laptop or desktop computer and it started to work. After the mother board sends a signal to the [power supply](http://en.wikipedia.org/wiki/Power_supply), the power supply provides the computer with the proper amount of electricity. Once motherboard receives the [power good signal](http://en.wikipedia.org/wiki/Power_good_signal), it tries to run the CPU. The CPU resets all leftover data in its registers and sets up predefined values for every register.
+Despite that this is a series of posts about linux kernel, we will not start at the kernel code. 
+
+So, first you pressed the magic power button on your laptop or desktop computer and the computer begins to turn on. After the mother board sends a signal to the [power supply](http://en.wikipedia.org/wiki/Power_supply), the power supply unit provides the computer with the proper amount of electricity. Once the motherboard receives the [power good signal](http://en.wikipedia.org/wiki/Power_good_signal), it attempts to run the CPU (Central Processing Unit). The CPU resets all leftover data in its registers and sets up predefined values for every register.
 
 
 [80386](http://en.wikipedia.org/wiki/Intel_80386) and later CPUs defines the following predefined data in CPU registers after the computer resets:
@@ -34,7 +36,9 @@ CS selector 0xf000
 CS base     0xffff0000
 ```
 
-The processor starts working in [real mode](http://en.wikipedia.org/wiki/Real_mode) now and we need to make a little retreat for understanding memory segmentation in this mode. Real mode is supported in all x86 compatible processors, from [8086](http://en.wikipedia.org/wiki/Intel_8086) to modern Intel 64bit CPUs. The 8086 processor had a 20 bit address bus, which means that it could work with 0-2^20 bytes address space (1 megabyte). But it only had 16 bit registers, and with 16 bit registers the maximum address is 2^16 or 0xffff (64 kilobytes). Memory segmentation was used to make use of all of the address space. All memory was divided into small, fixed-size segments of 65535 bytes, or 64 KB. Since we can not address memory behind 64 KB with 16 bit registers, another method to do it was devised. An address consists of two parts: the beginning address of the segment and the offset from the beginning of this segment. To get a physical address in memory, we need to multiply the segment part by 16 and add the offset part:
+The processor now starts working in [real mode](http://en.wikipedia.org/wiki/Real_mode) and we need to make a little retreat to understand memory segmentation in this mode. 
+
+Real mode is supported in all x86 compatible processors, from [8086](http://en.wikipedia.org/wiki/Intel_8086) to modern Intel 64bit CPUs. The 8086 processor had a 20 bit address bus, which means that it could work with 0-2^20 bytes address space (1 megabyte). But it only had 16 bit registers, and with 16 bit registers the maximum address is 2^16 or 0xffff (64 kilobytes). Memory segmentation was used to make use of all of the address space. All memory was divided into small, fixed-size segments of 65535 bytes, or 64 KB. Since we can not address memory behind 64 KB with 16 bit registers, another method to do it was devised. An address consists of two parts: the beginning address of the segment and the offset from the beginning of this segment. To get a physical address in memory, we need to multiply the segment part by 16 and add the offset part:
 
 ```
 PhysicalAddress = Segment * 16 + Offset
@@ -83,7 +87,7 @@ reset_vector:
 	...
 ```
 
-We can see here jump instruction [opcode](http://ref.x86asm.net/coder32.html#xE9) - 0xe9 to the address `_start - ( . + 2)`. And we can see that `reset` section is 16 bytes and starts at `0xfffffff0`:
+We can see here the jump instruction [opcode](http://ref.x86asm.net/coder32.html#xE9) - 0xe9 to the address `_start - ( . + 2)`. And we can see that `reset` section is 16 bytes and starts at `0xfffffff0`:
 
 ```
 SECTIONS {
@@ -97,7 +101,7 @@ SECTIONS {
 }
 ```
 
-Now BIOS has started to work. After all initializations and hardware checking, it needs to load operating system. BIOS tries to find bootable device which contains boot sector. Boot sector is the first sector on device (512 bytes) and contains sequence of `0x55` and `0xaa` at 511 and 512 byte. For example:
+Now BIOS has started to work. After all initializations and hardware checking, it needs to load the operating system. BIOS tries to find a bootable device which contains a boot sector. A boot sector is the first sector on the device (512 bytes) and contains a sequence of `0x55` and `0xaa` at 511 and 512 byte. For example:
 
 ```assembly
 [BITS 16]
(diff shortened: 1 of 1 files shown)


Huddayn:
In my opinion , it doesn't imporve readability. And it also adds semantic errors in some places.

Regards

Booting/linux-bootstrap-1.md
```diff
@@ -446,25 +450,25 @@ Ok now we have correct segment registers, stack, need only setup bss and jump to
...
-Jump to main
+Jump to Main
 --------------------------------------------------------------------------------
 
-That's all, we have stack, bss and now we can jump to `main` C function:
+That's all, we have the stack, bss and now we can jump to the `main` C function:
 
 ```assembly
-	calll main
+	calll  main
```
0xAX: drop space here please

Booting/linux-bootstrap-1.md
```diff
@@ -446,25 +450,25 @@ Ok now we have correct segment registers, stack, need only setup bss and jump to
 	rep; stosl
 ```
 
-First of all we put [__bss_start](https://github.com/torvalds/linux/blob/master/arch/x86/boot/setup.ld#L47) address in `di` and `_end + 3` (+3 - align to 4 bytes) in `cx`. Clear `eax` register with `xor` instruction and calculate size of BSS section (put in `cx`). Divide `cx` by 4 and repeat `cx` times `stosl` instruction which stores value of `eax` (it is zero) and increase `di`by the size of `eax`. In this way, we write zeros from `__bss_start` to `_end`:
+First of all we put [__bss_start](https://github.com/torvalds/linux/blob/master/arch/x86/boot/setup.ld#L47) address in `di` and `_end + 3` (+3 - align to 4 bytes) in `cx`. Next, we clear `eax` register with `xor` instruction and calculate the size of the BSS section (put in `cx`). Divide `cx` by 4 and repeat `cx` times `stosl` instruction which stores value of `eax` (it is zero) and increase `di` by the size of `eax`. In this way, we write zeros from `__bss_start` to `_end`:
 
 ![bss](http://oi59.tinypic.com/29m2eyr.jpg)
 
-Jump to main
+Jump to Main
```
0xAX: `main` must be in lowercase, here, because function name

Booting/linux-bootstrap-1.md
```diff
@@ -165,24 +169,24 @@ General real mode memory map is:
...
 ```
 
-At the start of execution BIOS is not in RAM, it is located in ROM.
+At the start of execution the BIOS is not in RAM, it is located in ROM.
 
 Bootloader
 
-Now BIOS has transferred control to the operating system bootloader and it needs to load operating system into the memory. There are a couple of bootloaders which can boot linux, like: [Grub2](http://www.gnu.org/software/grub/), [syslinux](http://www.syslinux.org/wiki/index.php/The_Syslinux_Project) and etc... Linux kernel has [Boot protocol](https://github.com/torvalds/linux/blob/master/Documentation/x86/boot.txt) which describes how to load linux kernel.
+Now the BIOS has transferred control to the operating system bootloader, it needs to load the operating system into memory. There are a number of bootloaders which can boot linux, including: [Grub2](http://www.gnu.org/software/grub/) and [syslinux](http://www.syslinux.org/wiki/index.php/The_Syslinux_Project). There is a linux [Boot protocol](https://github.com/torvalds/linux/blob/master/Documentation/x86/boot.txt) which describes how to load the linux kernel.
```
0xAX: number?

Booting/linux-bootstrap-1.md
```diff
@@ -130,26 +134,26 @@ We will see:
...
 
-**NOTE**: as you can read above the CPU is in real mode. In real mode, calculating the physical address in memory is as follows:
+**NOTE**: as you can see above, the CPU is in real mode. In real mode, calculating the physical address in memory is as follows:
 
 ```
 PhysicalAddress = Segment * 16 + Offset
 ```
 
-as I wrote above. But we have only 16 bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
+as we discussed above. But we have only 16 bit general purpose registers. The maximum value of 16 bit register is: `0xffff`; So if we take the biggest values, it will be:
```
0xAX: hmm, i'm not sure, but i think that `wrote` was better, what do you think?

Booting/linux-bootstrap-1.md
```diff
@@ -130,26 +134,26 @@ We will see:
...
+In this example we can see that this code will be executed in 16 bit real mode and will start at 0x7c00 in memory. After starting it calls the [0x10](http://www.ctyme.com/intr/rb-0106.htm) interrupt which just prints the `!` symbol. It fills the rest of the 510 bytes with zeros and finishes with two magic bytes 0xaa and 0x55.
 
-Real world boot loader starts at the same point, ends with `0xaa55` bytes, but reads kernel code from device, loads it to memory, parses and passes boot parameters to kernel and etc... instead of printing one symbol :) Ok, so, from this moment BIOS handed control to the operating system bootloader and we can go ahead.
+Real world boot loader starts at the same point, ends with `0xaa55` bytes, but reads kernel code from the device, loads it to memory, parses and passes boot parameters to kernel and etc... instead of printing one symbol :) Ok, so, from this moment on, BIOS has handed control to the operating system bootloader and we can proceed.
```
0xAX: why is 'on' here?

I mean: `so, from this moment on`

Booting/linux-bootstrap-1.md
```diff
@@ -1,29 +1,31 @@
...
-* Understanding C code
-* Understanding assembly code (AT&T syntax)
+* An understanding of C code
+* An understanding of assembly code (AT&T syntax)
 
-Anyway, if you just started to learn some tools, I will try to explain some parts during this and following posts. Ok, little introduction finished and now we can start to dive into kernel and low-level stuff.
+If you just started to learn some tools, I will try to explain some parts during this and the following posts. Ok, our little introduction is finished and now we can start to dive into th kernel and low-level stuff.
 
-All code is actual for kernel - 3.18, if there are changes, I will update posts.
+All code is actually for kernel - 3.18, if there are changes, I will update my posts.
```
0xAX: why 'actually' here? I think that 'actual' is more suitable, or i'm wrong?

Booting/linux-bootstrap-1.md
```diff
@@ -1,29 +1,31 @@
-Kernel booting process. Part 1.
+Kernel Booting Process. Part 1.
 ================================================================================
 
-From the bootloader to kernel
+From the Bootloader to the Kernel
```
0xAX, line 4: What's the purpose to make it uppercase?

0xAX:
@georgehorrell first of all thank you for your work. I have left some comments.

@Huddayn thank you for review of this PR too... I left some comments, maybe i have missed something else?

Booting/linux-bootstrap-1.md
```diff
@@ -302,9 +306,9 @@ _start:
 	//
 ```
 
-Here we can see `jmp` instruction opcode - `0xeb` to the `start_of_setup-1f` point. `Nf` notation means following: `2f` refers to the next local `2:` label. In our case it is label `1` which goes right after jump. It contains rest of setup [header](https://github.com/torvalds/linux/blob/master/Documentation/x86/boot.txt#L156) and right after setup header we can see `.entrytext` section which starts at `start_of_setup` label.
+Here we can see `jmp` instruction opcode - `0xeb` to the `start_of_setup-1f` point. `Nf` notation means following: `2f` refers to the next local `2:` label. In our case it is the label `1` which goes right after jump. It contains the rest of the setup [header](https://github.com/torvalds/linux/blob/master/Documentation/x86/boot.txt#L156) and right after the setup header we can see the `.entrytext` section which starts at `start_of_setup` label.
```
0xAX, line 309: @Huddayn for me too, although i'm not english-man, all these a/the looks like magic for me:)

Huddayn:
Ok, I've commented on the parts that I think should be changed.

georgehorrell:
Huddayn, those are all very reasonable points; I'm not hugely familiar with technical writing but do have some experience in copywriting. Sorry about the numerous typos. 

How do I submit my correction to all of the errors?

Booting/linux-bootstrap-1.md
```diff
@@ -1,29 +1,31 @@
...
 
 **Required knowledge**
 
-* Understanding C code
-* Understanding assembly code (AT&T syntax)
+* An understanding of C code
+* An understanding of assembly code (AT&T syntax)
 
-Anyway, if you just started to learn some tools, I will try to explain some parts during this and following posts. Ok, little introduction finished and now we can start to dive into kernel and low-level stuff.
+If you just started to learn some tools, I will try to explain some parts during this and the following posts. Ok, our little introduction is finished and now we can start to dive into th kernel and low-level stuff.
```
georgehorrell: > Anyway, if you just started

The colloquialisms broke my flow when I was reading it over - I'd hate to take away from 0xAX's style though. Just a judgement call. 

> this and the following

This is correct. 'Following' means _in response to_ or _as a result of_ whereas 'the following' is used to introduce a noun or collective noun.

> the kernel

will correct this

Booting/linux-bootstrap-1.md
```diff
@@ -423,19 +427,19 @@ If `CAN_USE_HEAP` bit is set, put `heap_end_ptr` to `dx` which points to `_end`
...
 ```assembly
 cmpl	$0x5a5aaa55, setup_sig
 jne	setup_bad
 ```
 
-just consists of comparing of [setup_sig](https://github.com/torvalds/linux/blob/master/arch/x86/boot/setup.ld#L39) and `0x5a5aaa55` number, and if they are not equal jump to error printing.
+just consists of comparing of [setup_sig](https://github.com/torvalds/linux/blob/master/arch/x86/boot/setup.ld#L39) and `0x5a5aaa55` number, and if they are not equal, jump to error printing.
 
-Ok now we have correct segment registers, stack, need only setup bss and jump to C code. Bss section used for storing statically allocated uninitialized data. Here is the code:
+Now we have correct segment registers, stack, we only to setup bss and jump to C code. Bss section used for storing statically allocated uninitialized data. Here is the code:
```
georgehorrell: will correct

Booting/linux-bootstrap-1.md
```diff
@@ -235,14 +239,14 @@ where `X` is the address kernel bootsector loaded. In my case `X` is `0x10000` (
...
 
-Start of kernel setup
+Start of Kernel Setup
 --------------------------------------------------------------------------------
 
-Finally we are in the kernel. Technically kernel didn't run yet, first of all we need to setup kernel, memory manager, process manager and etc... Kernel setup execution starts from [arch/x86/boot/header.S](https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S) at the [_start](https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S#L293). It is little strange at the first look, there are many instructions before it. Actually....
+Finally we are in the kernel. Technically, the kernel didn't run yet, first of all we need to setup the kernel, memory manager, process manager and etc... Kernel setup execution starts from [arch/x86/boot/header.S](https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S) at the [_start](https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S#L293). It is a little strange at first sight, as there are many instructions before it. But if we look back...
```
georgehorrell: > the kernel didn't run yet

also I think this should read

> the kernel hasn't run yet

Booting/linux-bootstrap-1.md
```diff
@@ -165,24 +169,24 @@ General real mode memory map is:
...
+Now the BIOS has transferred control to the operating system bootloader, it needs to load the operating system into memory. There are a number of bootloaders which can boot linux, including: [Grub2](http://www.gnu.org/software/grub/) and [syslinux](http://www.syslinux.org/wiki/index.php/The_Syslinux_Project). There is a linux [Boot protocol](https://github.com/torvalds/linux/blob/master/Documentation/x86/boot.txt) which describes how to load the linux kernel.
 
-Let us briefly consider how grub loads linux. GRUB2 execution starts from `grub-core/boot/i386/pc/boot.S`. It starts to load from device its own kernel (not to be confused with linux kernel) and executes `grub_main` after successfully loading.
+Let us briefly consider how grub loads linux. GRUB2 execution starts from `grub-core/boot/i386/pc/boot.S`. It starts by loading from the device, its own kernel (not to be confused with linux kernel) and then executes `grub_main`.
```
georgehorrell, line 185: will correct

Booting/linux-bootstrap-1.md
```diff
@@ -165,24 +169,24 @@ General real mode memory map is:
...
 
-But stop, at the beginning of post I wrote that first instruction executed by the CPU is located at address `0xfffffff0`, which is much bigger than `0xffff` (1MB). How can CPU access it in real mode? As I write about and you can read in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
+But wait a minute, at the beginning of the post I wrote that the first instruction executed by the CPU is located at address `0xfffffff0`, which is much bigger than `0xffff` (1MB). How can the CPU access it in real mode? As I write about and you can read in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
 
 ```
 0xFFFE_0000 - 0xFFFF_FFFF: 128 kilobyte ROM mapped into address space
 ```
 
-At the start of execution BIOS is not in RAM, it is located in ROM.
+At the start of execution the BIOS is not in RAM, it is located in ROM.
```
georgehorrell, line 178: will correct

Booting/linux-bootstrap-1.md
```diff
@@ -165,24 +169,24 @@ General real mode memory map is:
 0x000F0000 - 0x000FFFFF - System BIOS
 ```
 
-But stop, at the beginning of post I wrote that first instruction executed by the CPU is located at address `0xfffffff0`, which is much bigger than `0xffff` (1MB). How can CPU access it in real mode? As I write about and you can read in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
+But wait a minute, at the beginning of the post I wrote that the first instruction executed by the CPU is located at address `0xfffffff0`, which is much bigger than `0xffff` (1MB). How can the CPU access it in real mode? As I write about and you can read in [coreboot](http://www.coreboot.org/Developer_Manual/Memory_map) documentation:
```
georgehorrell, line 172: will correct

Booting/linux-bootstrap-1.md
georgehorrell: does `actual` have a technical meaning? This might be better as

> All code is written for kernel - 3.18

Booting/linux-bootstrap-1.md
georgehorrell: > There are couple of bootloaders which can boot linux, including: Grub2 and syslinux.

missing an 'a' before couple? I avoided couple because there are [more than two](http://en.wikipedia.org/wiki/Comparison_of_boot_loaders) bootloaders.

> Linux kernel has Boot protocol which describes how to load the linux kernel.

This doesn't make sense to me. In the documentation it is referred to as _the linux/x86 boot protocol_, your rephrasing would suggest that the kernel is loading itself.

Booting/linux-bootstrap-1.md
georgehorrell: will correct

Booting/linux-bootstrap-1.md
georgehorrell: will correct

Booting/linux-bootstrap-1.md
georgehorrell: stylistic preference. I will change it back.

Booting/linux-bootstrap-1.md
georgehorrell, line 4: Capitalising the nouns is normally how you format a title, if you would rather not I will change those back also.

Booting/linux-bootstrap-1.md
```diff
@@ -97,7 +101,7 @@ SECTIONS {
 }
 ```
 
-Now BIOS has started to work. After all initializations and hardware checking, it needs to load operating system. BIOS tries to find bootable device which contains boot sector. Boot sector is the first sector on device (512 bytes) and contains sequence of `0x55` and `0xaa` at 511 and 512 byte. For example:
+Now BIOS has started to work. After all initializations and hardware checking, it needs to load the operating system. BIOS tries to find a bootable device which contains a boot sector. A boot sector is the first sector on the device (512 bytes) and contains a sequence of `0x55` and `0xaa` at 511 and 512 byte. For example:
```
georgehorrell, line 104: I'm slowly getting more familiar with where to insert the articles in this type of documentation - thanks for correcting this for me.

Huddayn:
@georgehorrell Just make another commit to the same fork.

Booting/linux-bootstrap-1.md
ccope, line 104: BIOS isn't a proper noun, it's an acronym for a common noun (there are numerous BIOSes). Therefore, it should be referred to as 'the BIOS'.

Booting/linux-bootstrap-1.md
```diff
@@ -130,26 +134,26 @@ We will see:
 
 ![Simple bootloader which prints only `!`](http://oi60.tinypic.com/2qbwup0.jpg)
 
-In this example we can see that this code will be executed in 16 bit real mode and will start at 0x7c00 in memory. After the start it calls the [0x10](http://www.ctyme.com/intr/rb-0106.htm) interrupt which just prints `!` symbol. It fills rest of 510 bytes with zeros and finish with two magic bytes 0xaa and 0x55.
+In this example we can see that this code will be executed in 16 bit real mode and will start at 0x7c00 in memory. After starting it calls the [0x10](http://www.ctyme.com/intr/rb-0106.htm) interrupt which just prints the `!` symbol. It fills the rest of the 510 bytes with zeros and finishes with two magic bytes 0xaa and 0x55.
```
ccope, line 137: Instead of '0x7c00 in memory' I would write 'memory address `0x7c00`'

Booting/linux-bootstrap-1.md
ccope, line 172: CPU is an acronym for another common noun (central processing unit). The use of 'the' is correct.

Booting/linux-bootstrap-1.md
ccope, line 178: This 'the' is correct.

Booting/linux-bootstrap-1.md
ccope: The wording there is a bit awkward, `on` and `proceed` shouldn't be used together like this. `At this moment, the BIOS has handed control to the bootloader.` Please revert the change to this line though, I've updated the bootloader section in another pull request.

Booting/linux-bootstrap-1.md
ccope: There are a couple of ways to phrase this. I believe you originally intended to say `All kernel code in this document is actual code from Linux 3.18`. I think `This document contains sample code from Linux 3.18.` is better.

Booting/linux-bootstrap-1.md
ccope, line 104: I actually added some detail to this section in my pull request.

Booting/linux-bootstrap-1.md
ccope, line 185: I rewrote this section in my PR.

ccope:
@georgehorrell  if you're working in the web interface, go here and click the edit button: https://github.com/georgehorrell/linux-insides/blob/booting-grammar-fixes/Booting/linux-bootstrap-1.md
Then just commit your changes on the same branch, and they will show up here.

If you're on the command line, just make the changes, `git add` the filename, `git commit`, and `git push`.

0xAX:
@georgehorrell thank you that you updated PR.

I can't merge it without conflicts. Can you please rebase on must. If you will need any help with this, let me know

0xAX:
@caffinatedmonkey I know :) But It will be easier to review it. @georgehorrell please rebase on master and I will review and merge it.

Thank you

PR #56, opened 2015-02-11
Fix typo

GuillaumeGomez:
I think the 's' is too much.


Huddayn:
There is no such word ‘serie’, while ‘series’ can be *both* singular and plural.

GuillaumeGomez:
Well, I did learn something then !

PR #58, opened 2015-02-24
linux-bootstrap-1.md -- fixed some grammar errors

LYF610400210:


0xAX:
Thank you @LYF610400210, merged

LYF610400210:
@simar7 We cannot make full use of a memory that is larger than 64kB, since we cannot express an address that is >65535 with a 16-bit value, so we use two 16-bit values to express a big address.

simar7:
@LYF610400210 that makes sense. Thanks for the explanation. Maybe this could be rephrased as you suggested in your comment.

PR #60, opened 2015-02-28
Add description on QEMU booting

jonte:
This further explains that the snippet is boot sector code, and that we are instructing QEmu to load the code as a disk image.


0xAX:
@jonte Thank you. merged :+1:

PR #61, opened 2015-03-01
Make x86 paging reference title clearer

cirosantilli:
Knowing it is a tutorial makes it clearer what it is.

Google analytics ;)


0xAX:
thank you @cirosantilli

PR #64, opened 2015-03-09
fix typo: 1MB is 0xfffff, not 0xffff which is 65535

fabiopozzi:
fix typo in linux-bootstrap-1: 1MB is near 0xfffff, not 0xffff which is 65535


0xAX:
Thank you @fabiopozzi, merged

PR #65, opened 2015-03-09
Intel syntax uses ; as a comment mark

plashchynski:
There's a NASM compilation error:
```
boot.nasm:1: error: label or instruction expected at start of line
Note: this example written with Intel syntax:1: error: label or instruction expected at start of line
```


0xAX:
Thank you @plashchynski

PR #67, opened 2015-03-15
Fix Grammatical Errors

comjf:
A few Grammatical Errors. For future reference, don't use: 'and etc' the and isn't needed.

src: http://grammar.about.com/od/alightersideofwriting/a/Etc-And-Et-Al.htm

p.s. I love these articles, keep them coming!


0xAX:
Thank you @comjf for PR. Merged.

PR #73, opened 2015-03-22
missing letters added.

akash0x53:


0xAX:
Thank you again @akash0x53 !

Initialization/linux-initialization-3.md
```diff
@@ -181,7 +181,7 @@ if (paravirt_enabled())
 	return;
 ```
 
-we exit from the `reserve_ebda_region` function if paravirtualization is enabled because if it enabeld the extended bios data are is absent. In the next step we need to get the end of the low memory:
+we exit from the `reserve_ebda_region` function if paravirtualization is enabled because if it enabeled the extended bios data are is absent. In the next step we need to get the end of the low memory:
```
ihrwein, line 184: IMHO it's `enabled` and not `enabeled` :)

PR #79, opened 2015-03-23
Kevinjswinton bootstrap patch

kevinjswinton:
Not an x86 expert but the **jmp boot** seems unnecessary - a quick test suggests this is the case, so I thought I'd submit this simple change (I'm working on the basis that I am correct and it is unnecessary, in which case let's remove it so as not to confuse readers, or it's actually *necessary* - in which case it'll need an *explanation*!)


0xAX:
@kevinjswinton thank you. Yes jmp is not necessary here

PR #81, opened 2015-03-27
Update linux-bootstrap-1.md

kevinjswinton:
A small review of the BSS area. I hope you find the changes acceptable - the changes proposed here should improve the narrative flow, as well as provide a slightly clearer explanation as to what the x86 code is doing (for those less familiar with native x86 instructions).


0xAX:
Thank you @kevinjswinton

PR #91, opened 2015-04-14
Theory fixes

darkstar:
The "triple-backtick" quote used doesn't wrap correctly (shows horizontal scrollbar) and uses the wrong font (monospaced). Using the ">" syntax is better in that case.


0xAX:
Thank you @darkstar

PR #93, opened 2015-04-19
Fix nasm output format

arpankapoor:
Output option for flat-form binary files is 'bin'


0xAX:
Thank you @arpankapoor

PR #121, opened 2015-06-28
Update interrupts-1.md

noxiouz:
In #230 I only added `It *is* used` in the second sentence, but the diff looks weird.


TheCodeArtist:
Since it is a long line (word-wrapped into multiple on-screen lines) the diff view shows the entire line as-is except for your change.

Was this what you described as weird?

noxiouz:
Yep :)

0xAX:
Thank you @noxiouz

PR #131, opened 2015-07-16
Fix a few english typos - "can be runned not"

diogovk:


0xAX:
Thank you @diogovk

Agares:
Shouldn't it be "Cannot be ran"/"Can't be ran"?

diogovk:
You're right. There's no such thing as runned. 
The participle of run is run as well. 
So I think the correct would be cannot be run.

PR #141, opened 2015-07-28
It should be "cld" here :)

nimisolo:
... It should be "cld" here :)


0xAX:
Ah, yes, thank you.

Agares:
It should be `cli`, `cld` clears direction flag.

mudongliang:
I have the same doubt. 
 - cld is used to clear DF in movXX instruction. 
 - cli is used to disable the interrupts

PR #180, opened 2015-08-24
Typo "an" to "a"

ubershmekel:


anubiann00b:
I don't understand why this is so hard to comprehend. I'm looking at this trying to convince myself that it should be "an," but it makes no sense in my head :laughing:

epheph:
Might as well fix "an user" later in the paragraph while you're at it.

0xAX:
So... `a userspace` and `an user program`?

epheph:
No, both are "a", but there is an "an" 12 words prior to end of paragraph.

epheph:
This PR makes the paragraph more correct, it is fine to merge. There's just another instance of the same error later in the paragraph, but it could be fixed separately.

rpeiremans:
The rule isn't the letter following a or an, it's rather the sound of that letter. When the word start with a vowel sound you use an, otherwise it's a. For example an hour, since you pronounce the h as a vowel, but a user because the u sounds like a consonant in this case.

http://www.englishpage.com/articles/a-vs-an.htm has quite a clear explanation as well, including some more examples.

epheph:
@rpeiremans  yes, that's what i'm campaigning for. There's just two mistakes on this line and this PR only fixes one.

0xAX:
Thank all, merged

PR #201, opened 2015-09-12
Fix spelling error

tscanlon:
@0xAX thanks for your work.

The github UI doesn't show the one character change so here it is in summary `applyes` -> `applies`


0xAX:
Thank you @tscanlon, merged

PR #204, opened 2015-09-13
"and etc" => "etc"

joar:
"etc." stands for "et cetera", "et" means "and".

Sidenote: & also stands for "et", so "etc." may be written as "&c.".


0xAX:
Thank you @joar

PR #207, opened 2015-09-14
Fix a small typo "learnd" -> "learned".

hristozov:
This PR fixes the typo in:

> We started from the understanding of the `system call` concept, we have **learnd** what is it and...


0xAX:
Thank you @hristozov

PR #224, opened 2015-09-22
Fixed some grammatical errors.

stormbeard:


stormbeard:
This will be on-going as I find them. I'm working from the end backwards to the beginning since there may be some other folks helping with this.

0xAX:
Thank you @stormbeard

PR #227, opened 2015-09-22
Recommend ivandaviov/minimal to generate initrd

cirosantilli:
Script that does what you already say, but in a single bash command, really convenient.

I also wanted to recommend my fork https://github.com/cirosantilli/runlinux which allows to do it from any kernel tree (declined upstream https://github.com/ivandavidov/minimal/issues/20), but I was afraid that would be too much self promotion :-)

### DIFF
--- a/Misc/contribute.md
+++ b/Misc/contribute.md
@@ -277,6 +277,8 @@ $ qemu-system-x86_64 -snapshot -m 8GB -serial stdio -kernel ~/dev/linux/arch/x86
 
 From now we can run the Linux kernel in the virtual machine and this means that we can begin to change and test the kernel.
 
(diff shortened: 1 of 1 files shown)


0xAX:
Hello @cirosantilli, please use markdown:

    Consider using [minimal](https://github.com/ivandavidov/minimal) to automate the process of generating initrd.

cirosantilli:
How about `ivandaviov/minimal`? Minimal is too generic.

But I still thing that the clearest possibility is to put the full URL as it becomes clear to people that it is a GitHub repo.

`<http://a.com>` is markdown :-) http://spec.commonmark.org/0.21/#autolinks

0xAX:
@cirosantilli Looks good, Thank you, merged

PR #231, opened 2015-09-24
Slight grammatical cleanup of first 3 paragraphs

lengau:
I attempted to keep the same personal feel in the two paragraphs, but modified their content to sound more 'normal' to native English speakers.

I'm learning more from this than I can contribute on a technical level, but I'd like to help in what ways I can.


0xAX:
Thank you

PR #238, opened 2015-10-05
fix small syntax problems, add new methods for list traverse in the last part

mudongliang:


0xAX:
Thank you @mudongliang

PR #249, opened 2015-10-12
fixed grammar in linux-bootstrap-5.md, Preparation before kernel decompression section

ruthgrace:

### DIFF
--- a/Booting/linux-bootstrap-5.md
+++ b/Booting/linux-bootstrap-5.md
@@ -9,7 +9,7 @@ This is the fifth part of the `Kernel booting process` series. We saw transition
 Preparation before kernel decompression
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


Booting/linux-bootstrap-5.md
```diff
@@ -38,9 +38,9 @@ ENTRY(startup_64)
 	movl	%eax, %gs
 ```
 
-in the beginning of the `startup_64`. All segment registers besides `cs` points now to the `ds` which is `0x18` (if you don't understand why it is `0x18`, read the previous part).
+in the beginning of the `startup_64`. All segment registers besides `cs` now point to the `ds` which is `0x18` (if you don't understand why it is `0x18`, read the previous part).
```
0xAX: Why `point` here instead `points`?

0xAX:
@ruthgrace Hello, First of all thank you for PRs. This looks very good for me and I've merged it. Have no much time, will look on other PRs later.

PR #252, opened 2015-10-14
Explaining what are packages and why are they useful

gustavofulton:


mudongliang:
Why do you add this concept? I think this repo focus on the details of Linux kernel, other than details of Linux distribution. Package and package manager , like deb,rpm; apt-get,rpm,zypper,pacman,yaourt e.g. , is not necessary to be included.

0xAX:
Hello @gabesullice,

@mudongliang is fully right here, this repo is only related to the Linux kernel, so I'm closing this

PR #258, opened 2015-10-14
fixed grammar in linux-bootstrap-4.md, Calculate relocation address section

ruthgrace:

### DIFF
--- a/Booting/linux-bootstrap-4.md
+++ b/Booting/linux-bootstrap-4.md
@@ -221,7 +221,7 @@ We set stack, checked CPU and now can move on the next step.
 Calculate relocation address
 --------------------------------------------------------------------------------
 
-The next step is calculating relocation address for decompression if need. We can see following assembly code:
+The next step is calculating relocation address for decompression if needed. We can see the following assembly code:
 
 ```assembly
 #ifdef CONFIG_RELOCATABLE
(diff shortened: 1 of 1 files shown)


Booting/linux-bootstrap-4.md
```diff
@@ -262,22 +262,22 @@ Let's look on the code. If we have `CONFIG_RELOCATABLE=n` in our kernel configur
...
+Here we calculate the aligned address where the kernel is loaded (`0x100000` or 1 megabyte in our case). `PHYSICAL_ALIGN` is an alignment value to which the kernel should be aligned, and it ranges from `0x200000` to `0x1000000` for x86_64. With the default values we will get 2 megabytes in the `LOAD_PHYSICAL_ADDR`:
 
 ```python
 >>> 0x100000 + (0x200000 - 1) & ~(0x200000 - 1)
 2097152
 ```
 
-After that we got alignment unit, we adds `z_extract_offset` (which is `0xe5c000` in my case) to the 2 megabytes. In the end we will get 17154048 byte offset. You can find `z_extract_offset` in the `arch/x86/boot/compressed/piggy.S`. This file generated in compile time by [mkpiggy](https://github.com/torvalds/linux/blob/master/arch/x86/boot/compressed/mkpiggy.c) program.
+After we retrieve the alignment unit, we add `z_extract_offset` (which is `0xe5c000` in my case) to the 2 megabytes. In the end we will get 17154048 bytes offset. You can find `z_extract_offset` in `arch/x86/boot/compressed/piggy.S`. This file is generated in compile time by the [mkpiggy](https://github.com/torvalds/linux/blob/master/arch/x86/boot/compressed/mkpiggy.c) program.
```
0xAX, line 272: Maybe `retrieved` here?

ruthgrace, line 272: @0xAX both "After we retrieve..." and "After we have retrieved..." are correct, and mean the same thing. Do you have any preference?

0xAX, line 272: It was just little note, this is good for me if this is correct. Thank you

PR #264, opened 2015-10-16
revert "internals" to "insides"

mudongliang:


mudongliang:
I have a little similar problem : There are some other "internals" string in the markdown file. But they are not for url, but like "Book about Linux kernel internals", "try to understand internals of these and other time management concepts" e.g.
I don't know whether to modify "internals" to "insides" in them. Neither do I know why you change name of repo.

0xAX:
Hello @mudongliang, I think we must rename all to `linux-insides`.

mudongliang:
OK, I will then pull requests to revise all the rest "internal" to "insides". Thanks!

PR #268, opened 2015-10-18
fixed grammar in linux-bootstrap-3.md, Heap API section

ruthgrace:

### DIFF
--- a/Booting/linux-bootstrap-3.md
+++ b/Booting/linux-bootstrap-3.md
@@ -58,7 +58,7 @@ If you read source code of the kernel, you'll see these very often and so it wil
 Heap API
 --------------------------------------------------------------------------------
 
-After we have `vid_mode` from the `boot_params.hdr` in the `set_video` function we can see call to `RESET_HEAP` function. `RESET_HEAP` is a macro which defined in the [boot.h](https://github.com/torvalds/linux/blob/master/arch/x86/boot/boot.h#L199). It is defined as:
+After we have `vid_mode` from `boot_params.hdr` in the `set_video` function we can see a call to the `RESET_HEAP` function. `RESET_HEAP` is a macro which is defined in [boot.h](https://github.com/torvalds/linux/blob/master/arch/x86/boot/boot.h#L199). It is defined as:
 
 ```C
 #define RESET_HEAP() ((void *)( HEAP = _end ))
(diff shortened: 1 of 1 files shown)


Booting/linux-bootstrap-3.md
```diff
@@ -58,7 +58,7 @@ If you read source code of the kernel, you'll see these very often and so it wil
 Heap API
 --------------------------------------------------------------------------------
 
-After we have `vid_mode` from the `boot_params.hdr` in the `set_video` function we can see call to `RESET_HEAP` function. `RESET_HEAP` is a macro which defined in the [boot.h](https://github.com/torvalds/linux/blob/master/arch/x86/boot/boot.h#L199). It is defined as:
+After we have `vid_mode` from `boot_params.hdr` in the `set_video` function we can see a call to the `RESET_HEAP` function. `RESET_HEAP` is a macro which is defined in [boot.h](https://github.com/torvalds/linux/blob/master/arch/x86/boot/boot.h#L199). It is defined as:
```
0xAX: Can we fix:

After we have `vid_mode` from `boot_params.hdr` in the `set_video` function we can see a call to the `RESET_HEAP`

As

After we have got the `vid_mode` from the `boot_params.hdr` in the `set_video` function, we can see the call of the `RESET_HEAP`

ruthgrace: Hi @0xAX,
Just to clarify a bit about my thought process in editing this:

1) For proper nouns (i.e. names) like `boot_params.hdr`, you would leave out the word 'the', but if it's a common noun like 'register' or 'function', you would say "the register" or "the function", even if there is an adjective, like "the `RESET_HEAP` function". You can see a really long but thorough explanation here: http://www.trussel.com/the.htm

2) "Call to the `RESET_HEAP` function" means that RESET_HEAP is being called, whereas "call of the `RESET_HEAP` function" means that RESET_HEAP is calling you (or the current function you're describing). Another example of this English phrasing is the game "Call of Duty", which means that duty is calling you (to fight in the war).

So I think to preserve your meaning but to keep it grammatically correct, the sentence should be:
"After we get `vid_mode` from `boot_params.hdr` in the `set_video` function, we can see the call to the `RESET_HEAP` function."

There are a couple other options for the first part: "After we retrieve `vid_mode`...", "After we have gotten `vid_mode`...", "After we retrieve the value of `vid_mode`...", and "After we get the value of `vid_mode`...", "After we have gotten the value of `vid_mode`...". Let me know which you prefer. :)

Ruth

0xAX:
Hello @ruthgrace thank you for your explanation. Yes this:

`After we get vid_mode from boot_params.hdr in the set_video function, we can see the call to the RESET_HEAP function`

variant is very good.

ruthgrace:
@0xAX PR has been updated :)

0xAX:
@ruthgrace Thank you!

PR #280, opened 2015-10-26
revert "internals" to "insides" in most parts 

mudongliang:


mudongliang:
There are only Booting part which has the internals. Once the PR related to Booting is merged, I will pull requests for the last fix.

0xAX:
Thank you @mudongliang

PR #299, opened 2015-11-02
revert internals to insides in Booting

mudongliang:
This is the last part of changing. "git grep" will not find any "internals" in the repo.


0xAX:
Thank you @mudongliang

PR #301, opened 2015-11-11
Fix 'calll' typo in booting

MartinBrugnara:
Cleaned version of #300


0xAX:
It is not a typo. It is a long call - https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S#L509

MartinBrugnara:
Sorry my fault!

PR #309, opened 2015-12-31
fix statements errors and grammer errors in kernel initialization

hitmoon:
@0xAX should be better!


hitmoon:
@0xAX I have do the best I can to fix logic and grammer errors in this chapter, hope it will be help!

0xAX:
@hitmoon many thanks

PR #315, opened 2016-01-11
fix interrupts/interrupts-4

spacewander:


0xAX:
Thank you @spacewander

PR #321, opened 2016-01-19
fix errors in mm/

hitmoon:


hitmoon:
@0xAX fix mm

0xAX:
Thank you @hitmoon

PR #327, opened 2016-01-31
call typo fix

8tab:


0xAX:
It does not error. Its a long `call`

PR #335, opened 2016-03-14
timers-2.md: typo fix, dependend -> depdendent

pclouds:
Side note, I find "specific" to  be easier to understand in these contexts, e.g. "x86-specific timers". But I'm not native English speaker either :)


0xAX:
Thank you @pclouds

PR #350, opened 2016-03-27
Fix typo

nemesis:


nemesis:
Didn't notice one was already open

PR #364, opened 2016-04-02
bootstrap-1: help readers not to confuse segment and physical address

relip:
I was a bit confused while reading linux-bootstrap-1, like 0x1000 suddenly became 0x10000 and 0x200 became 0x20 etc.

This might help readers not to confuse them at least this part, but still, more explanation of memory addressing will be more helpful for the beginners!


0xAX:
Thank you @relip

PR #366, opened 2016-04-03
Fix few typos.

abhinav-upadhyay:
s/function which is will be called during initialization of the Linux kernel of right after/function which will be called during initialization of the Linux kernel or right after

s/Acutally/Actually


0xAX:
Thank you @abhinav-upadhyay

PR #370, opened 2016-04-11
Fix small grammatical error

nehaljwani:
s/use/uses . The word 'processor' is singular :)


0xAX:
Thank you @nehaljwani

PR #375, opened 2016-04-18
Correct the include/linux/spinlock.h URL

thismachinechills:
The URL specified for the file `include/linux/spinlock.h` is incorrect. The correct link is [https://github.com/torvalds/linux/blob/master/include/linux/spinlock.h](https://github.com/torvalds/linux/blob/master/include/linux/spinlock.h).


0xAX:
Thank you @thismachinechills

PR #377, opened 2016-04-18
Fix MCA link

Alcaro:
Links are \[text\](URL) in Markdown, not (text)[URL].


Alcaro:
More fixes incoming, I found about a dozen typos and unusual wordings in linux-bootstrap-4. I'll publish them when I'm done reading that chapter.

0xAX:
Thank you @Alcaro

PR #378, opened 2016-04-18
Fix a huge pile of issues

Alcaro:
You've done a good job with this book, despite a considerable language barrier. Here's my attempt at clearing up the language issues, and a couple of technical issues I found as well.

I've only read the boot chapter thus far. I'll submit more PRs for the other chapters once I get there.


0xAX:
Thank you @Alcaro

PR #385, opened 2016-05-02
Fix some mixed types in asm.md.

Smattr:
This commit fixes some instances where return types did not match the type of the variable being returned or `printf` format strings did not match the types of arguments.

In the case of the return types of functions, the types silently coerce so I don't think it's a big deal, though it makes for examples that are less clear than they could be. In the case of the `printf` format argument, I think this is actually incorrect as an `int` is 4 bytes and an `unsigned long` is 8 bytes on x86_64. In practice everything's probably fine because I believe the `int` gets implicitly extended to 8 bytes by being passed through a register (as in http://blog.regehr.org/archives/1303), but I would not think this is safe in general. I realise this is all a bit IMHO and nitpicky, so if you disagree feel free to refuse this pull request.

Thanks for your work in this repo :)


0xAX:
Thank you @Smattr

PR #388, opened 2016-05-03
Fix typo in syscall-2.md

yunomae:
`rxc` may be `rcx`, so we fix it.


0xAX:
@yunomae sorry for this long delay and thank you!

PR #390, opened 2016-05-08
Fix some missing code block end markers

thanatos:
Some of the code blocks in `SyncPrim/sync-4.md` were missing their end markers,
causing normal text to be added to the code block. Fix that.


0xAX:
Thank you @thanatos

PR #391, opened 2016-05-09
Fix malformed links in SyncPrim 3 & 4 and Initialization 7.

JGRennison:
'CPU mask' link in Initialization 7 is double-bracketed.
'Previous part' links in SyncPrim 3 & 4 have two opening brackets
and one closing.
'Previous part' link in SyncPrim 4 points to current page,
instead of previous.


0xAX:
Thank you @JGRennison

PR #405, opened 2016-06-05
Correct a typo calll -> call

d33tah:


0xAX:
@d33tah Thank you, but `calll` is a long call instruction

d33tah:
@0xAX: ah, sorry about that then. Thanks for clarifying :)

PR #407, opened 2016-06-21
<Theory/ELF: Fix a link to kernel memory map documentation>

bobsaintcool:
The link pointed to "EFI region mapping space" instead of "kernel text
mapping, from phys 0".

I think that the link to the mm documentation point to the wrong line.


0xAX:
Thank you @bobsaintcool

PR #412, opened 2016-07-10
Fix typos and incorrect explanation in seqlock theory paragraphs

mullen3:
This commit fixes a few typos and also an incorrect explanation of read behavior in the [Sequential Lock theory section](https://github.com/0xAX/linux-insides/blob/master/SyncPrim/sync-6.md#sequential-lock). 

I believe the explanation of the different scenarios for the reader in this section is incorrect. It originally stated that the counter values being *NOT* equal after executing a critical section indicated that there hadn't been any writers. As previously explained, the writers increment the counter value before execution, which makes this explanation incorrect. I also attempted to improve to the explanation of the scenarios.


0xAX:
Many thanks @mullen3

PR #417, opened 2016-07-25
Grammar and typofixes

Shugyousha:
The first commit contains a few grammar and typo fixes. The second one actually changes the defines to the enum values that are specified as the memblock flags.

So the second commit warrants a closer look.


0xAX:
Thank you @Shugyousha

PR #427, opened 2016-09-11
Fix userspace startup process typos.

cloudburst:
Correct various typos in "program startup process in userspace"


0xAX:
Thank you @cloudburst

PR #434, opened 2016-10-27
change filename to match naming style

mudongliang:


0xAX:
Thank you @mudongliang

PR #436, opened 2016-10-29
fix typos and semantics in program startup

mudongliang:


0xAX:
Thank you @mudongliang

PR #450, opened 2017-01-07
Change README links as relative

proninyaroslav:
Hello. This will make the universal navigation not only on GitHub, but also on GitBook and in PDF files.


proninyaroslav:
Hi, that I can fix in this patch that it be merged?

0xAX:
@proninyaroslav Thank you for pr and sorry for delay

PR #459, opened 2017-02-19
Update asm.md

walkerlala:
maybe typos

### DIFF
--- a/Theory/asm.md
+++ b/Theory/asm.md
@@ -159,7 +159,7 @@ As mentioned above, the "clobbered" part should contain a comma-separated list o
 Consider the example from before, but we will add an additional, simple assembler instruction:
 
 ```C
-__asm__("movq $100, %%rdx\t\n"
+__asm__("movq $100, %%edx\t\n"
         "addl %1,%2" : "=r" (sum) : "r" (a), "0" (b));
 ```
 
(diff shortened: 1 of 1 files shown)


mudongliang:
Why do you modify "rax/rdx" to "eax/edx"? The "rax/rdx" is 64bit register.
Refer to [https://wiki.cdot.senecacollege.ca/wiki/X86_64_Register_and_Instruction_Quick_Start#General-Purpose_Registers](https://wiki.cdot.senecacollege.ca/wiki/X86_64_Register_and_Instruction_Quick_Start#General-Purpose_Registers)

walkerlala:
sorry I didn't spot that. But I suppose mixing 64-bit codes and 32-bit code is not a good idea. It is easier to get confused when with that (and note that the machine code generated have no indication of how many bits the CPU is)

mudongliang:
Yes, you're right. But there are still 64bit registers in this file, such as "rsp", you need to modify them all, or  discuss with @0xAX

0xAX:
Hello everyone,

@walkerlala first of all thanks for PR.

> But I suppose mixing 64-bit codes and 32-bit code is not a good idea

Absolutely. But this code is generated by gcc. Maybe yes, it will be good idea to rewrite it with 32-bit registers, but in this way we all need to edit operands like `movq` to 32-bit variants.

walkerlala:
Well, as the document is used for memo and illustration rather than production use, I would still suggest change them all to 32-bit format to make things clean. That would only involve some changes of registers. I have create a new pull request.

I have not actually compile my code and replace the gcc-generated machine code, but rather hand-craft some of them. Nevertheless, I think that is sufficient considering the purpose of this document.

Thanks.

jcrben:
Out of curiosity, why go with 32-bit rather than 64-bit? If I'm playing around with the examples and learning assembly I feel like I'd rather use and get familiar with 64-bit registers. It does sound like there might be more complexity with QEMU and 64-bit tho, especially on Windows...

walkerlala:
On Sat, May 27, 2017 at 05:23:06PM -0700, Ben Creasy wrote:
> Out of curiosity, why go with 32-bit rather than 64-bit? If I'm playing around with the examples and learning assembly I feel like I'd rather use and get familiar with 64-bit registers. It does sound like there might be more complexity with QEMU and 64-bit tho, especially on Windows...
 
Indeed, pure 64-bit and pure 32-bit would be great. The point is, the current
example mix 32-bit and 64-bit instructions (that is the output of GCC)

0xAX:
should be fixed in the https://github.com/0xAX/linux-insides/commit/896411ad172b7d6cc8ec501980049cec4538f483

PR #460, opened 2017-02-21
Update asm.md

walkerlala:
fix mixed 32-bit/64-bit codes


mudongliang:
It looks that you have already modified all the x86_64 registers. Please double check content. I see some paras are related with x86_64 arch.

0xAX:
I've updated the `Theory/asm.md` part in the https://github.com/0xAX/linux-insides/commit/896411ad172b7d6cc8ec501980049cec4538f483 and now all examples are remade with 64-bit registers. 

I'm going to close this and the prior pr. If you have any additional issues let me know.

Thank you

PR #482, opened 2017-05-29
Remove dead link

palimm:
The domain systemcomputing.org seems to be registered but there seems to be no dns
configuration. So the sites is not available.


0xAX:
👍

PR #483, opened 2017-05-29
Corrected typos in section: memory

phantez:
> again to to cause nest step page fault. -> again to cause next step to page fault.
> it must to map to - > it must map it to


0xAX:
Thank you @phantez

PR #486, opened 2017-06-06
correct minor mistake - list 3 bits instead of 4

TomasJuocepis:
Sentence starts with "The next three bits(40, 41, 42, 43)" Should exclude 43.


0xAX:
Thank you @TomasJuocepis 👍

PR #488, opened 2017-06-12
fix:fix the wrong words x87 to x86.

woodpenker:
We don't have X87,it's x86.
[Ref to this issue](https://github.com/0xAX/linux-insides/issues/487)


0xAX:
@woodpenker thanks for PR, but it is not error. See http://wiki.osdev.org/Exceptions#x87_Floating-Point_Exception and https://en.wikipedia.org/wiki/X87. So I'm closing this PR.

woodpenker:
My Mistake.Thank U

PR #490, opened 2017-06-20
Update README with brazilian portuguese translation

mauri870:
I'll work on the translation in my spare time. Great job with this posts and keep with your awesome work.


0xAX:
@mauri870 great 👍

PR #492, opened 2017-07-11
Update the 'kernel parameter' link address

gerryqd:
The old address is outdated, it is updated by the new one.


0xAX:
Thank you @gerryqd

PR #495, opened 2017-07-17
Typo fixes on booting part1

litao91:
As in our case, the kernel is loaded at 0x10000, the setup code should start from `0x10200` instead of `0x1020`, I believe that this is a typo.


0xAX:
ah yes, it was wrong fix, thank you @litao91!

PR #499, opened 2017-07-30
SysCall/syscall-6: Fix typos

jwilk:
I didn't fix this sentence:

> Such resources are not infinite and each process and we should have an instrument to manage it.

I can't parse it, but I also don't know what you wanted to say here. Perhaps "and each process" should be removed?


0xAX:
@jwilk thank you

PR #500, opened 2017-07-30
Update sentence grammar

int3rrupt:
Remove "never" from end of sentence and change "can't exceed to "can never exceed"


0xAX:
Thank you @int3rrupt

PR #507, opened 2017-08-10
Clarify: vDSO is linked automatically with glibc

diekmann:
Change is partly a verbatim copy of https://www.kernel.org/doc/Documentation/ABI/stable/vdso

Quote: "Programs that dynamically link to glibc will use the vDSO automatically. Otherwise, you can use the reference parser in tools/testing/selftests/vDSO/parse_vdso.c."

The old version of the book was slightly imprecise. It was unclear to me whether this is an optional or a default behavior.

The main semantic change is the word 'automatically'.


0xAX:
👍

PR #508, opened 2017-08-10
Language: Added articles. Content: exec* family

diekmann:
Language improvement: Added articles like `the`, `a`, ...

Content: The `exec*` family is a `C` wrapper, the syscall is usually only `execve`.


0xAX:
Thank you @diekmann 👍

PR #515, opened 2017-09-05
Rdo corrections

dasty:
Thank you for all the great work - two patches fixes two minor issues noticed when reading your book.


0xAX:
@dasty 👍

PR #519, opened 2017-09-15
Fix register typos and correct explanation of the first ljmpl target

danix800:
addl %ebx, 2f => add value of ebx to memory labeled '2', to construct the correct jump target.


0xAX:
👍

PR #539, opened 2017-11-03
fix calll to call

tkmru:


0xAX:
Hello @tkmru, thank you for PR but this is not a typo. See https://github.com/torvalds/linux/blob/master/arch/x86/boot/header.S#L618

tkmru:
oh, sorry

PR #551, opened 2017-11-29
Fix a description on phys_base 

firogh:
In the following expression, we add the value of phys_base to the
rax register, not the address of phys_base.
addq    phys_base(%rip), %rax

BTW, the value of phys_base =
the actual loaded physical address of kernel after relocation - the
likned physical address of kernel.

Signed-off-by: Firo Yang <firogh@gmail.com>


0xAX:
👍

PR #554, opened 2017-12-29
Fixed mistakes in the segment descriptors part

ayzenquwe:
When I started reading the description of segment descriptors it was really hard to understand "the general scheme" and where limit/base/attributes are located. So I decided to search for other descriptions/tutorials, and I understood why it was so hard for me - there were small, but very confusing mistakes.

PR fixes those mistakes and makes "the general scheme" a little clearer by removing relative bits numbers.


0xAX:
@ayzenquwe oh thank you 👍 merged

PR #563, opened 2018-01-27
Stack address fix

0xF0D0:
Sorry for the messy commits. I'm currently working on translation for hobby

but I'm no really well-trained GitHub user.

Anyway, at stack setup code,
`cmpw    %ax, %dx`

%ax has address of %ds,

so I thought the valid value should be `0x10000` not `0x1000`


0xAX:
Hello @0xF0D0, thank you for your PR. 

But I think you're not right, and the address is correct. See [Memory addressing in real mode](https://wiki.osdev.org/Real_Mode#Memory_Addressing). So, in our case

`0x1000 * 16` is `0x10000` which is base address of the kernel setup code.

0xF0D0:
Opps, I've got confused between base address and Segment Register. 

Sorry.

I think it's ok to close.

PR #567, opened 2018-02-14
Adopted standard filename convention and automated LaTeX conversion/build.

ranguli:
I was interested in reading this book in a LaTex format, so I've created a script that will automatically do the following:

- It goes through each subdirectory, converts the Markdown files into a singular LaTex file for that chapter.
- Collects each chapter file into the build directory.
- Compiles them into PDFs.
- Merges the PDF together.

The script lives in the project root as `latex.sh` and all the processed files are in `build` although it is set to be git ignored. It's arguably quite rudimentary so I'd be receptive to improving it. Currently it can't deal with any images, for example. Bit of a bummer, I know.

I needed them to have a similar filename convention so I could glob for them and compile them (which I went and implemented)  but I don't see this as a downside as most of the files were already following a `linux-chaptername-1-md` convention. 

My sort of use-case is that if you're reading it at a computer or online, GitBooks is the way to go (it's really nice.) But if you wanted to print it off and have a hard copy now you can do so with an equally pretty version. 

Let me know if you have any issues.

Cheers
Josh


0xAX:
Hello @ranguli, nice work.

> My sort of use-case is that if you're reading it at a computer or online, GitBooks is the way to go (it's really nice.) But if you wanted to print it off and have a hard copy now you can do so with an equally pretty version.

Yes, but GtiBook already provides PDF version [here](https://www.gitbook.com/book/0xax/linux-insides/details)

ranguli:
Hi there, I was aware of that, but I think it's still an extremely handy option for those who prefer the look of LaTeX documents for viewing.

0xAX:
Hello @ranguli,

> Hi there, I was aware of that, but I think it's still an extremely handy option for those who prefer the look of LaTeX documents for viewing.

Yes, that's OK. Only one last question:

> I needed them to have a similar filename convention so I could glob for them and compile them (which I went and implemented) but I don't see this as a downside as most of the files were already following a linux-chaptername-1-md convention.

Is it just `*.md` not enough? Can we avoid files renaming to preserve git  history/blame?

ranguli:
`*.md` isn't enough because the script needs to identify `README.md` first and then the files of that chapter so that in the exported PDF the `README` shows up first, then the chapter contents.

It was my impresson that renaming files won't delete their history/blame. Plus, naming conventions are always good to have

0xAX:
Hello @ranguli, sorry me for such long delay but it is merged now. Thank you for pr :+1:

mudongliang:
We may need to make the PR open for now.
Since you change the filename of those markdown file, the URLs related with those renamed files should be changed. 

Maybe we need a small script to handle this job.

mudongliang:
If you guys don't mind, could you please discuss the filename conversion in issue #568 ?

dato:
Thanks for providing the initial conversion output. Much more important now that Gitbook is not providing PDF?

Unfortunately, it doesn't create the book in order, nor does it have a table of contents. (And Preface.tex is not being converted to PDF.)

I do not have time to improve the script; but I wanted to leave this one-liner (that fixes the three mentioned issues) in case it's useful to somebody. From the top-level directory (must `apt install librsvg2-bin` first):

```
pandoc --pdf-engine=xelatex README.md SUMMARY.md CONTRIBUTING.md $(sed -rne 's/.*\((.*\.md).*/\1/p' SUMMARY.md) --resource-path=$(ls -d */ | tr '\n' :). -o LinuxInsides.pdf
```

This is brittle (for example, it breaks if there are images with the same name in different sub-directories)—but it worked for me.

PR #580, opened 2018-03-20
check and modify all the links that contains interrupts

mudongliang:


mudongliang:
@0xAX , if you don't have enough time, you could leave the fix of links to me at first.

0xAX:
Thanks for this @mudongliang

PR #581, opened 2018-03-20
 move latex.sh to Scripts folder

mudongliang:


mudongliang:
@0xAX I modify the content of latex.sh to make it suitable in the Scripts folder. You could have a test of this script when you have installed some proper linux packages.

PR #583, opened 2018-03-23
 Finish fixing broken links related with Theory, Data structure, Misc and so on

mudongliang:
First, I have almost finished fixing all the broken links with keyword - "0xax". There is only 1 broken link with keyword "0xax" in the repo - http://0xax.github.io/categories/assembly/. I don't know how to modify it.

Then for some "http" links in the gitbook, I may miss. However, I could write one script to modify all the "http://0xax.gitbook.io/..." to "https://0xax.gitbook.io/..." we both prefer.

Finally, all the things will return back to normal.


0xAX:
Hello @mudongliang,

> First, I have almost finished fixing all the broken links with keyword - "0xax". There is only 1 broken link with keyword "0xax" in the repo - http://0xax.github.io/categories/assembly/. I don't know how to modify it.

Seems it just should be `https`. Will fix it after this PR.

> Then for some "http" links in the gitbook, I may miss. However, I could write one script to modify all the "http://0xax.gitbook.io/..." to "https://0xax.gitbook.io/..." we both prefer.

Yes it should be pretty easy to do. I will try to find time to do it on these weekends.

> Finally, all the things will return back to normal.
:+1: :+1: :+1:

0xAX:
Now it should be fixed in the last commit https://github.com/0xAX/linux-insides/commit/264d9101826946ee391922e812a2840433acc836

0xAX:
@mudongliang btw maybe we should add travis-ci or something like this to run the script after commits to catch wrong/dead links? 

What do you think about this?

mudongliang:
Good idea.

PR #598, opened 2018-06-18
Update linux-interrupts-9.md

ramosian-glider:
Fix style/grammar nits I've ran into while reading about interrupts.


0xAX:
Thank you @ramosian-glider

PR #602, opened 2018-07-11
Add kernelhacking mailing list

hao-lee:
Hi, I create a Google Group mailing list to discuss the latest kernel source code. linux-insides is a good starting point to dive into the kernel code, so I think we can use this mailing list as a place to learn kernel together. Mailing list is more convenient than Issues when discuss problems and Issues can be used for ebook contribution.


0xAX:
:+1:

PR #631, opened 2018-10-28
fix typo of "closk"

snyh:


mudongliang:
@snyh Thanks for your commit.

Issue #645, opened 2018-12-28
sentence needs to be improved


PinoTsao:
In "Kernel booting process. Part 4.", section "Reload the segments if needed", one sentence should be improved:

>The value assigned to it is an offset relative to the offset of the segment

==>

>The value assigned to it is an offset relative to the segment

tkyymmt:
Would you mind sending PR?

PinoTsao:
yes, but I need to learn how to sending a PR, please wait for a while:)

PinoTsao:
> After comparison of the LOAD_PHYSICAL_ADDR and value of the ebx register, we add the offset from the startup_32 where to decompress the compressed kernel image. If the CONFIG_RELOCATABLE option is not enabled during kernel configuration, we just put the default address where to load kernel and add z_extract_offset to it.

To be honest, I think most content of it is wrong. It is just a calculation of physical address where to place the decompressed kernel, if we don't have CONFIG_RELOCATABLE, it is the LOAD_PHYSICAL_ADDR, or else, it will be the address of %ebp(protect mode kernel loaded address) aligned to BP_kernel_alignment, while no less than LOAD_PHYSICAL_ADDR

PinoTsao:
>The compressed kernel image should be moved to the end of the decompression buffer to simplify calculations where kernel will be located later. For this:

Although there is no obvious buffer ending address in the code, but I feel "moved against to the end xxx" is more accurate, "against" has more meaning that the compressed kernel wouldn't exceed the buffer.  And I don't quite understand why the latter part exists, how about:

>The compressed kernel image will be moved against the end of the decompression buffer for safety in-place decompression.

PinoTsao:
> setting 5 bit in it and loading it again into cr4

==>

> setting the 5th bit in it and loading it again into cr4

PinoTsao:
> > After comparison of the LOAD_PHYSICAL_ADDR and value of the ebx register, we add the offset from the startup_32 where to decompress the compressed kernel image. If the CONFIG_RELOCATABLE option is not enabled during kernel configuration, we just put the default address where to load kernel and add z_extract_offset to it.
> 
> To be honest, I think most content of it is wrong, for example, I don't see any reference of z_extract_offset in current code here, maybe in older time it is different.
> 
[...]

PR #648, opened 2019-01-15
Sentences improvement in linux-bootstrap-4.md

PinoTsao:
This patch fixed half of issue of #645:
1.
"The value assigned to it is an offset relative to the offset of the segment"
-->
"The value assigned to it is an offset relative to the segment"

2.
"adding the address field of the instruction and the value of the program counter"
-->
"adding the address field of the instruction to the value of the program counter"

3.
"PAE mode" --> "PAE", because PAE is not a mode.

4.
"5 bit" --> "the 5th bit"

### DIFF
--- a/Booting/linux-bootstrap-4.md
+++ b/Booting/linux-bootstrap-4.md
@@ -127,7 +127,7 @@ SECTIONS
 }
 ```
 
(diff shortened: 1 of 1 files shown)


0xAX:
Thank you @PinoTsao :tada:

PR #651, opened 2019-02-19
cand to and

RorraVox:
"From this moment we have `argc`cand `argv" -> "From this moment we have `argc` and `argv"


mudongliang:
@RorraVox thanks

PR #663, opened 2019-03-17
Proposed improvements to explanation of spinlock

tlkh:
Only modified the paragraph on spinlock. Added more elaboration and tried to explain a bit clearer. Do let me know what you think!


0xAX:
Looks good to me, thank you @tlkh

PR #687, opened 2019-09-23
Corrected escaping on two URLs

md1512:
Corrected escaping on two Wikipedia URLs


mudongliang:
It seems your PR breaks the semantic of markdown. Please double check your commit.

PR #693, opened 2019-10-10
Fix (almost) all images by storing them locally

darkstar:
Tinypic has taken them all down by now, I could get all but one from the [WaybackMachine](https://web.archive.org)

This PR adds them into local `images/` subdirectories and fixes the links in the Markdown files


0xAX:
@darkstar Thank you very much

PR #712, opened 2020-04-21
Linux interrupts 1 update

initBasti:
Finished the update of the code passages to the recent state.
I have added links in the commit description, that lead to the commit of the change.
@0xAX If you don't like/need proof for the change please notify me.
Additionally, I have made small improvements to the sentence structure of 2 descriptions.

I would highly appreciate any feedback about the pull request!

### DIFF
--- a/Interrupts/linux-interrupts-1.md
+++ b/Interrupts/linux-interrupts-1.md
@@ -237,7 +237,7 @@ extern gate_desc idt_table[];
 ```
 
 where `gate_struct` is defined as:
-[/arch/x86/include/asm/desc\_defs.h](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/desc_defs.h)
+[/arch/x86/include/asm/desc_defs.h](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/desc_defs.h)
 
 ```C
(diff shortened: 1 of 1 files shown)


Interrupts/linux-interrupts-1.md
```diff
@@ -420,35 +422,35 @@ or
...
 
-where `&nmi` and `&double_fault` are addresses of the entries to the given interrupt handlers:
-
-```C
-asmlinkage void nmi(void);
-asmlinkage void double_fault(void);
-```
-
-defined in the [arch/x86/kernel/entry_64.S](https://github.com/torvalds/linux/blob/16f73eb02d7e1765ccab3d2018e0bd98eb93d973/arch/x86/kernel/entry_64.S)
+where `nmi` and `double_fault` are entry points created at [arch/x86/kernel/entry\_64.S](https://github.com/torvalds/linux/blob/master/arch/x86/entry/entry_64.S):
```
0xAX: why is the `\` symbol here in the filename?

initBasti: I placed those because I thought they need to be escaped for the markdown syntax. Sorry for the mistake, I have added an additional commit to fix the problem.

PR #714, opened 2020-04-29
Remove empty entries from SUMMARY.md

initBasti:
I noticed this, while checking the different parts for the gitbook.
remove list entries from the summary to cause less confusion.
@0xAX Do you want to keep the empty entries as placeholders for future parts?


0xAX:
>  @0xAX Do you want to keep the empty entries as placeholders for future parts?
Yes, but still don't have time to finish them

PR #723, opened 2020-06-24
Update linux-interrupts-5.md: improve the grammar

GIHOLee:
improve the grammar in that case:
s/SIGFPE for the divide error, SIGILL for the overflow exception and etc.../
SIGFPE for a divide error, SIGILL for a invalid instruction etc.../


GIHOLee:
@initBasti Thanks, I was negligent.

PR #729, opened 2020-08-25
clock_getres is added to vDSO for x86 as of Jun 22, 2019

iAnatoly:
clock_getress is added to vDSO for x86 as of Jun 22, 2019: see https://github.com/torvalds/linux/commit/f66501dc53e72079045a6a17e023b41316ede220

Man page for vdso also lists four vDSO for x86, but they are wrong. I've dumped vDSO as binary from my current system, and it is really there, I am not going mad :)


initBasti (approved):
Thanks a lot!

PR #731, opened 2020-09-12
fix typo: &lock-val to &lock->val

dragonly:
fix typo in kernel code examples.


initBasti (approved):
NIce catch thank you!

Issue #498, opened 2017-07-25
Why disk image needs to have origin set to 0x7c00?


slo:
Regarding sentence after ASM code for MBR: "the origin is set to 0x7c00 and we end with the magic sequence". I couldn't find any indication why boot binary should have org set to 0x7c00. I've only found information about BIOS, which copies MBR under this address. But I don't understand why image should have this origin. From my tests this directive seems to be irrelevant to this image, because qemu uses disk image, which doesn't have any origin. I've also compared both images(wtih and without directive org 0x7c00), and they are the same.

aniket-deole:
https://www.glamenv-septzen.net/en/view/6 might clear your doubt?

slo:
Maybe I was not clear. My question was only pointing out, that directive: [ORG  0x7c00] is irrelevant in context of binary output from masm (and during my investigation whether nasm source contained it or not, it produced the same binary output). So my conclusion is that [ORG  0x7c00] in disk image is not needed because it comes from running software, rather than device. And based on that, should be removed from disk source in article.

0xAX:
> So my conclusion is that [ORG 0x7c00] in disk image is not needed because it comes from running software, rather than device.

Ah yes, now it is clear. You're absolutelly right. I will update the part with a 'bootloader' code.

0xAX:
here is it https://github.com/0xAX/linux-insides/commit/abf4f684a5e20033bcf657babba1c77025292227

mpetch:
`ORG 0x7c00` is generally needed (with binary file output), but your code example is too simple to manifest the problem. In the absence of an ORG  (same as `ORG 0x0000`) It is true that QEMU and BIOSes will load the 512 byte bootloader at physical address 0x07c00, but the assembler has no knowledge of this. In your code example the instructions are all position independent. You can run them from anywhere in memory and it will still work. Because of this fact you can remove the `org 0x7c00` and generate the identical code.

Now what happens if you want to reference data in the the bootloader. Let us say rather than coding the `!` as part of a mov instruction we place the `!` elsewhere in the bootloader and load it from memory. Something like this:

```
;
; Note: this example is written in Intel Assembly syntax
;
[BITS 16]

boot:
    xor ax, ax                 ; AX = 0
    mov ds, ax                 ; Initialize DS=0. BIOS may not have set it to 0

    mov al, [charmem]          ; [charmem] is relative to DS, same as saying
                               ;     ds:[charmem] which is why DS must be set.
[...]

PR #734, opened 2020-09-26
linux-bootstrap-1: Clearify the boot sector origin example

roykuper13:
Hey! First of all thanks for providing this great content!

This commit follows abf4f68:

1. The sentence "fulfills the requirements of the boot sector (the origin is set to 0x7c00" confused me as the example code doesn't set the origin. After finding this issue (#498) I understood that the line that sets the origin was removed.

2. I actually wasn't sure why the origin should be set to 0x7c00. After reading online, i figured it out, so i thought it is also good to mention it (and also mention that it doesn't needed when providing a boot binary to qemu).

not sure if my diff is clear and/or accurate enough though, so i would like to know if rephrasing is needed. Thanks!

closes https://github.com/0xAX/linux-insides/issues/732, https://github.com/0xAX/linux-insides/issues/635


0xAX:
@roykuper13 thank you fro the PR :+1:

PR #749, opened 2020-11-04
Make sentences more readable

BDHU:


initBasti (changes requested):
Hey,

Thank you for the pull request. I found a few more things I would like you to address before we can merge it into master.

The '/' is part of a path not part of a name, therefore I propose the following change:
s/the name of the our binary file starts from the `/` symbol/the path of the binary file starts in the root ('/') directory/

As you use the word: 'whether, I expect atleast a second possibility, therefore I would start the part for a path relative to working directory like this:
s/or the path of the given executable binary is interpreted relative to the current working directory of the calling process
/or if the path of the specified executable binary is relative to the current working directory of the calling process

And finally I think this sounds better:
s/read above about this/read about this above/
or maybe even:
s/read above about this/read above for more information about this/

Greetings,
Sebastian

initBasti:
Hey @BDHU ,
do you still have an interest in working on this pull request? If that is not the case I will close this pull request soon.
Greetings,
Sebastian

BDHU:
@initBasti I agree your suggestion makes better sense.

BDHU:
@initBasti I agree with your proposal. Seems like them made more senses.

BDHU (commented):
@initBasti I agree with your proposal. Seems like them made more senses.

0xAX:
As since no update I'll close the PR, please feel free to re-open it

PR #752, opened 2020-11-16
fix: The segment selector of  IDT

chaffz:
The selector is a 16 bit value and must point to a valid descriptor in your GDT.

### DIFF
--- a/Initialization/linux-initialization-2.md
+++ b/Initialization/linux-initialization-2.md
@@ -115,7 +115,11 @@ Where:
 * `Offset` - is offset to entry point of an interrupt handler;
 * `DPL` -    Descriptor Privilege Level;
 * `P` -      Segment Present flag;
-* `Segment selector` - a code segment selector in GDT or LDT
+* `Segment selector` - a code segment selector in GDT or LDT (actually it always point to GDT in linux)
+```C
+#define __KERNEL_CS	(GDT_ENTRY_KERNEL_CS*8) // 0000 0000 0001 0000
+#define GDT_ENTRY_KERNEL_CS 2
+```
 * `IST` -    provides ability to switch to a new stack for interrupts handling.
 
 And the last `Type` field describes type of the `IDT` entry. There are three different kinds of gates for interrupts:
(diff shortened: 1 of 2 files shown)


Mutated1994:
I agree with @chaffz 

Kernel: v3.10-rc1
Arch:    x86-64

`set_intr_gate` is defined in the[ arch/x86/include/asm/desc.h](https://elixir.bootlin.com/linux/v3.10-rc1/source/arch/x86/include/asm/desc.h#L342) like the following:
```C
static inline void set_intr_gate(unsigned int n, void *addr)
{
	BUG_ON((unsigned)n > 0xFF);
	_set_gate(n, GATE_INTERRUPT, addr, 0, 0, __KERNEL_CS);
}
```
The `_set_gate` function is defined in the same file as the `set_intr_gate` function like the following:
```C
static inline void _set_gate(int gate, unsigned type, void *addr,
			     unsigned dpl, unsigned ist, unsigned seg)
{
	gate_desc s;

	pack_gate(&s, type, (unsigned long)addr, dpl, ist, seg);
	/*
	 * does not need to be atomic because it is only done once at
	 * setup time
	 */
	write_idt_entry(idt_table, gate, &s);
}
```
The `pack_gate` function is also defined in the same file like the following:
```C
#ifdef CONFIG_X86_64

static inline void pack_gate(gate_desc *gate, unsigned type, unsigned long func,
			     unsigned dpl, unsigned ist, unsigned seg)
{
	...
	// Let's look at it.
	gate->segment		= __KERNEL_CS; 
	...
}

#else
```
The `__KERNEL_CS` is declared in the [arch/x86/include/asm/segment.h](https://elixir.bootlin.com/linux/v3.10-rc1/source/arch/x86/include/asm/segment.h#L187) and looks like:
```C
#define __KERNEL_CS	(GDT_ENTRY_KERNEL_CS*8) // 0000 0000 0001 0000
#define GDT_ENTRY_KERNEL_CS 2                         

15                                                 3    2        0
+--------------------------------------------------+----+--------+
|          Index                                   | TI |   RPL  |
+--------------------------------------------------+----+--------+

TI = Table Indicator: 0 = GDT, 1 = LDT
```
So, the selector points to a valid descriptor in GDT.
@0xAX @initBasti

PR #754, opened 2020-11-17
feat: Fill in the details of __PAGE_OFFSET

Mutated1994:


Mutated1994:
I‘m trying to submit a PR once per day.

KEEP GOING!

initBasti (changes requested):
Hey, thank you for the patch.
I would change the following sentences:

s/
The `early_recursion_flag` locates in the same assembly file as the `early_idt_handler_common` symbol like the following:/
The `early_recursion_flag` is defined in the same assembly file as the `early_idt_handler_common` symbol as follows:/

s/
and you can note that the `UL` tells the compiler that they are not of type int, but unsigned long./
The suffix `UL` forces the page offset to be a unsigned long data type./

Mutated1994:
> Hey, thank you for the patch.
> I would change the following sentences:
> 
> s/
> The `early_recursion_flag` locates in the same assembly file as the `early_idt_handler_common` symbol like the following:/
> The `early_recursion_flag` is defined in the same assembly file as the `early_idt_handler_common` symbol as follows:/
> 
> s/
> and you can note that the `UL` tells the compiler that they are not of type int, but unsigned long./
> The suffix `UL` forces the page offset to be a unsigned long data type./

Gotcha!

initBasti (approved):
Cool thanks for the fast adjustment.

Mutated1994:
> Cool thanks for the fast adjustment.

It's my pleasure.

PR #755, opened 2020-11-18
feat: Diving a little deeper about the PAGE_OFFSET

Mutated1994:


Mutated1994:
Fighting!

initBasti:
Hey @Mutated1994 ,
Thanks for the patch, one thing I noticed about your pull requests is that you barely provide any description for your changes. Those small additions can greatly improve the ability of other developers to comprehend the 'why' of a certain change. For example, in this patch, you add an additional explanation and it would help me for the review and future readers to understand why you felt that this is necessary.
Greetings,
Sebastian

initBasti (changes requested):
s/Then, let's dive a little deeper. From/Let us examine that address within the Documentation at/

The address 0xffff880000000000 is not mentioned anymore within the recent documentation. So if we add an explanation to this part, we should first make sure that the information is up to date. Could you please update this part so that it refers to the correct PAGE_OFFSET?

Initialization/linux-initialization-2.md
```diff
@@ -442,9 +442,8 @@ And the `_AC` macro is defined in the [include/uapi/linux/const.h](https://elixi
 #define _AC(X,Y)	__AC(X,Y)
 #endif
 ```
-So, where `__PAGE_OFFSET` expands to `0xffff880000000000`.
+Where `__PAGE_OFFSET` expands to `0xffff880000000000`. But, why a virtual address can be translated to the physical address by simply subtracting `__PAGE_OFFSET`? The answer is in the [Documentation/x86/x86_64/mm.txt](https://elixir.bootlin.com/linux/v3.10-rc1/source/Documentation/x86/x86_64/mm.txt#L9) documentation:
```
initBasti: s/But, why a virtual address can be translated to the physical address by simply subtracting `__PAGE_OFFSET`?
 /But, why is it possible to translate a virtual address to a physical address by subtracting `__PAGE_OFFSET`?

Initialization/linux-initialization-2.md
```diff
@@ -454,9 +453,9 @@ ffff880000000000 - ffffc7ffffffffff (=64 TB) direct mapping of all phys. memory
 ...
 ```
 
-As you may see,  the virtual address space `ffff880000000000-ffffc7ffffffffff` is direct mapping of all physical memory. When the kernel wants to access all physical memory, it uses direct mapping.
+As explained above, the virtual address space `ffff880000000000-ffffc7ffffffffff` is direct mapping of all physical memory. When the kernel wants to access all physical memory, it uses direct mapping.
 
-We initialize `pmd` and pass it to the `__early_make_pgtable` function along with `address`. The `__early_make_pgtable` function is defined in the same file as the `early_make_pgtable` function as the following:
+Okay, let's get back to discussing `early_make_pgtable`. We initialize `pmd` and pass it to the `__early_make_pgtable` function along with `address`. The `__early_make_pgtable` function is defined in the same file as the `early_make_pgtable` function as the following:
```
initBasti: s/defined in the same file as the `early_make_pgtable` function as the following
 /defined in the same file as the `early_make_pgtable` function as follows/

In order to sound less repetitive.

Initialization/linux-initialization-2.md
```diff
@@ -454,9 +453,9 @@ ffff880000000000 - ffffc7ffffffffff (=64 TB) direct mapping of all phys. memory
 ...
 ```
 
-As you may see,  the virtual address space `ffff880000000000-ffffc7ffffffffff` is direct mapping of all physical memory. When the kernel wants to access all physical memory, it uses direct mapping.
+As explained above, the virtual address space `ffff880000000000-ffffc7ffffffffff` is direct mapping of all physical memory. When the kernel wants to access all physical memory, it uses direct mapping.
```
initBasti: As mentioned in another comment, those addresses are not up-to-date.

initBasti (changes requested):
Thanks for the additions, I added a few nitpicks and requests.

Mutated1994:
> s/Then, let's dive a little deeper. From/Let us examine that address within the Documentation at/
> 
> The address 0xffff880000000000 is not mentioned anymore within the recent documentation. So if we add an explanation to this part, we should first make sure that the information is up to date. Could you please update this part so that it refers to the correct PAGE_OFFSET?

Yep, gotcha.

Initialization/linux-initialization-2.md
ioworker0: > Thanks for the additions, I added a few nitpicks and requests.

Thank you for your review and feedback.

initBasti (approved):

Issue #758, opened 2020-11-22
[Help] linux-bootstrap


Mutated1994:
![kernel_first_address](https://raw.githubusercontent.com/0xAX/linux-insides/master/Booting/images/kernel_first_address.png)

I have no idea of how to get this memory dump in real mode.
I searched a lot on the internet but i can't find answers.

Any help is much appreciated. Thanks!

Mutated1994:
OMG, I got it !!!

Mutated1994:
Version: v4.14.207
Arch:     x86_64
```
root@parallels-vm:/usr/src/linux-4.14.207# more arch/x86/kernel/vmlinux.lds
...
SECTIONS
{
 . = (0xffffffff80000000 + ALIGN(0x1000000, 0x200000));
 phys_startup_64 = ABSOLUTE(startup_64 - 0xffffffff80000000);
 .text : AT(ADDR(.text) - 0xffffffff80000000) {
  _text = .;
  _stext = .;
....
```
``` sh
root@parallels-vm:/usr/src/linux-4.14.207# nm vmlinux|grep startup_64
0000000001000000 A phys_startup_64
ffffffff81000030 T secondary_startup_64
ffffffff810001f0 T __startup_64
ffffffff81000000 T startup_64
```
Here we can see the memory address of the entry point, which is `0x0000000001000000`. Let's go ahead.

Before trying to debug the kernel, please see [Booting a Custom Linux Kernel in QEMU and Debugging It With GDB](http://nickdesaulniers.github.io/blog/2018/10/24/booting-a-custom-linux-kernel-in-qemu-and-debugging-it-with-gdb/)

### Step 1
Booting in QEMU
```
qemu-system-x86_64 -kernel /usr/src/linux-4.14.207/arch/x86_64/boot/bzImage -nographic -append "console=ttyS0 nokaslr" -initrd /data/busybox/busybox-1.28.0/initramfs.cpio.gz -S -s
```
### Step 2
Attaching GDB to QEMU
```
gdb vmlinux
(gdb) target remote :1234
(gdb) hbreak *0x0000000001000000
(gdb) c
(gdb) dump binary memory /tmp/dump 0x0000 0x20000
```
### Step 3
```
root@parallels-vm:/tmp# hd /tmp/dump |grep -A 31 MZ
00010000  4d 5a ea 07 00 c0 07 8c  c8 8e d8 8e c0 8e d0 31  |MZ.............1|
[...]

PR #762, opened 2020-11-29
feat: Get a memory dump in real mode

Mutated1994:
#758 

I hope this can help others who have no idea of how to get this memory dump in real mode.

### DIFF
--- a/Booting/linux-bootstrap-1.md
+++ b/Booting/linux-bootstrap-1.md
@@ -249,7 +249,7 @@ When the bootloader transfers control to the kernel, it starts at:
 X + sizeof(KernelBootSector) + 1
 ```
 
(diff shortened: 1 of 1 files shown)


0xAX:
Hello @Mutated1994, probably the issue is not the best place to refer, maybe it would be better to put your findings from the issue in the post itself?

Mutated1994:
> Hello @Mutated1994, probably the issue is not the best place to refer, maybe it would be better to put your findings from the issue in the post itself?

Gotcha.

Mutated1994:
> Hello @Mutated1994, probably the issue is not the best place to refer, maybe it would be better to put your findings from the issue in the post itself?

 I have put my findings from the issue in the post itself.

PR #773, opened 2021-05-04
Update linux-bootstrap-5.md

h4child:


initBasti (commented):
Hmmm, I don't like the fact that it prompts directly to a download, and also the repository states that the repository is now maintained here: [on gitlab](https://gitlab.com/x86-psABIs/x86-64-ABI/-/tree/master) 
I would prefer to link to the new repository, the downside is that the user has to do the following:
```
sudo apt-get install texlive-latex-base texlive-fonts-recommended texlive-fonts-extra texlive-latex-extra
git clone https://gitlab.com/x86-psABIs/x86-64-ABI.git
cd x86-64-ABI
make pdf
```

But the upside is that the sources stay up-to-date and the link doesn't instantly prompt a download.

But I don't have a strong opinion about this maybe we let @0xAX decide.

h4child:
Yes, I thought of just putting the link to the gitlab repository, but it would not be similar to the one with only the download link.
I didn't put a gitlab link because it would have more information on how to build the pdf and there would be no space. Although it is also not difficult to rebuild using gitlab.
I should have put the gitlab link, because it is the source always updated.

h4child:
sorry.
I deleted the branch patch from my fork and didn't even realize that I ended up closing the pull

JaeTone (approved):

PR #777, opened 2021-05-24
datastructures-2: typo fix

pangteypiyush:


DataStructures/linux-datastructures-2.md
```diff
@@ -159,7 +159,7 @@ The first `radix_tree_insert` function takes three parameters:
 
 The `radix_tree_delete` function takes the same set of parameters as the `radix_tree_insert`, but without data.
 
-The search in a radix tree implemented in two ways:
+The search in a radix tree implemented in three ways:
```
initBasti: While we are at it let's fix a few obvious sentence structure errors right away.
s/The search in a radix tree implemented in three ways:/
   Searching through a radix tree is implemented in three ways:/

initBasti (changes requested):
Thank you for the fix, the change is obviously correct. I would just like to make sure that whenever we touch a sentence, we don't leave other errors hanging.

PR #788, opened 2021-09-06
Fix an image link in linux-misc-1.md

henrybear327:
The old `make menuconfig` image link is defunct now. Replacing it with a new link.


0xAX:
@henrybear327 thank you

PR #796, opened 2022-02-10
fix the C bit number from 43 to 42

jihot2000:
I am a reader, not a linux kernel expert. I guess the number of C bit should be 42 according to the scheme of the segment descriptor.


luohuang:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已经收到。
 
谢谢！

0xAX:
`42` is correct here. It could be or direction bit or confirming bit, so I am closing the PR

PR #798, opened 2022-03-28
Remove incorrect commas

md2perpe:
Those commas seem misplaced. Without them the parameter list makes more sense.

### DIFF
--- a/SysCall/linux-syscall-4.md
+++ b/SysCall/linux-syscall-4.md
@@ -77,9 +77,9 @@ We saw preparation before a system call called by a user application and after a
 
 ```
 SYSCALL_DEFINE3(execve,
-		const char __user *, filename,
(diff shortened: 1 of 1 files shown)


luohuang:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已经收到。
 
谢谢！

md2perpe:
There's no need to be nasty. I was blinded by the lines looking like declarations and missed that each line consisted of two separate arguments to a macro. Sorry for disturbing.

tijko:
Didn't mean to come off that way, I guess my inner Torvald took over.  No harm no foul, happy hacking!

tijko:
@luohuang I would move to close as this is wrong and just based off a mis-understanding.

0xAX:
so I am closing this one :)

PR #800, opened 2022-05-02
capitalize Linux everywhere

renaudgermain:


0xAX:
@renaudgermain thanks for you great work. As you may see I don't have much time to keep it in actual state and review PRs as soon as possible so if you have wish and ability drop me e-mail I can grant you permissions to push directly to repo.

PR #818, opened 2022-08-03
copyedit: misc chapter

renaudgermain:


renaudgermain:
@0xAX Last 2 PRs. I won't need write access to the repo, I'll be moving on after those. Just thought I could contribute a little (I find typos are breaking my "reading flow", so that might be useful to others as well).

0xAX:
@renaudgermain in any cases thank you very much

PR #837, opened 2023-08-24
Create a release with ebooks on push to master

siddhpant:
Use GitHub actions to create a release with PDF, EPUB, and MOBI files automatically on push to master. The action can also be triggered manually.

The commit ID will be used as tag for releases.

This action does not use GitBook, but a fork of it named HonKit. gitbook-cli isn't maintained since 6 years, whereas honkit is a maintained fork.

### DIFF
--- a/.github/workflows/generate_release.yaml
+++ b/.github/workflows/generate_release.yaml
@@ -1,0 +1,80 @@
+# Note: Indentation here is of 2 spaces.
+
+name: Generate release with ebooks
+
+
+on:
+  push:
+    branches:
+      - master
+
+  workflow_dispatch: {}  # For manual switch.
+# End of on.
+
+
+jobs:
+  release_ebooks:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+        with:
+          ref: master
+      
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "latest"
+
+      # HonKit is a maintained fork of GitBook.
+      - name: Install honkit
+        run: npm install honkit -g
+
+      # APT install can take a long time due to installing dependencies,
+      # so we will use caching.
+      - name: Install calibre for ebook-convert
+        uses: awalsh128/cache-apt-pkgs-action@latest
+        with:
+          packages: calibre
+
+      # HonKit uses highlight.js for syntax highlighting, which doesn't have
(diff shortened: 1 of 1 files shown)


mudongliang:
@0xAX this seems an interesting feature: Github action is useful to do such things, but I am not familiar with its syntax and semantic.

wendajiang:
Any progress at this thread?

siddhpant:
@wendajiang it's the gitbook cli which is deprecated. Gitbook moved everything to its own closed source online platform, which has auto sync with repo.

Though yes, an open-source solution like mdbook or honkit would be nice IMHO. Does mdbook support generation of ebooks (ideally without much overhead and changes)?

wendajiang:
> @wendajiang it's the gitbook cli which is deprecated. Gitbook moved everything to its own closed source online platform, which has auto sync with repo.
> 
> Though yes, an open-source solution like mdbook or honkit would be nice IMHO. Does mdbook support generation of ebooks (ideally without much overhead and changes)? Asking because only one system should be used to generate both website and ebooks.

Sure, mdbook has some unofficial plugin to generate epub or pdf file. I'm not familiar with node ecology, sorry for the mistake.

siddhpant:
> Sure, mdbook has some unofficial plugin to generate epub or pdf file.

Nice. Let's wait for maitainers input.

> I'm not familiar with node ecology, sorry for the mistake.

Dw, neither am I.

mudongliang:
I tried Honkit recently, and it works after revising assembly syntax. I prefer to use Honkit.

mudongliang:
And I prefer to revise assembly to x86asm before applying this PR. However, x86asmatt is not applied now.

siddhpant:
> And I prefer to revise assembly to x86asm before applying this PR. However, x86asmatt is not applied now.

I agree that would be great, but that will break the instance hosted at the gitbook website.

To avoid that, there should be a complete move from gitbook website to a static website made using honkit on GitHub pages. 

This sounds like a bigger task, hence I just used `sed` before generating the pdf.

mudongliang:
> > And I prefer to revise assembly to x86asm before applying this PR. However, x86asmatt is not applied now.
> 
> I agree that would be great, but that will break the instance hosted at the gitbook website.

👍 Let's finish the infrastructure movement and then change the keyword assembly.

> 
> To avoid that, there should be a complete move from gitbook website to a static website made using honkit on GitHub pages.
> 
> This sounds like a bigger task, hence I just used `sed` before generating the pdf.

PR #844, opened 2024-03-12
Remove redundant section header in linux-bootstrap-4.md

ericlinqq:
Remove the section header "Reload the segments if needed" as it is duplicated.

### DIFF
--- a/Booting/linux-bootstrap-4.md
+++ b/Booting/linux-bootstrap-4.md
@@ -319,9 +319,6 @@ it has been loaded at and the compile time physical address
 
 Now that we know where to start, let's get to it.
 
-Reload the segments if needed
---------------------------------------------------------------------------------
-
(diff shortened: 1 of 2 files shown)


luohuang:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已经收到。
 
谢谢！

mudongliang:
I don't think we need to remove this section header.

mjbear:
> I don't think we need to remove this section header.

There are two identical section headers.

Should they both have the same name?

https://github.com/0xAX/linux-insides/blob/2eefbdfed90c30ce3bd1e89577b9957b5102dc25/Booting/linux-bootstrap-4.md?plain=1#L96

https://github.com/0xAX/linux-insides/blob/2eefbdfed90c30ce3bd1e89577b9957b5102dc25/Booting/linux-bootstrap-4.md?plain=1#L322

luohuang:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已经收到。
 
谢谢！

0xAX:
Since this chapter was reworked, it should be not relevant anymore, so I am closing the PR.

PR #849, opened 2024-06-02
replace(Initialization/linux-initialization-1.md): sme link l:135

vctorized:
replaced what looked like a wrong link for SME in the first chapter on kernel initialisation


mudongliang:
https://de.wikipedia.org seems a little wired. If you change de to en, you will jump to the original link. So I don't think we need to chagne this link

luohuang:
这是来自QQ邮箱的自动回复邮件。
 
您好，您的邮件我已经收到。
 
谢谢！

PR #870, opened 2025-08-25
Add book.json descriptor for gitbook exporter

kyselejsyrecek:
**Description**

Changes proposed in this pull request:

Without further configuration files present in the repository, gitbook fails to find the book title and author. Furthermore, it defaults to the A4 format which is inappropriate for e-book readers. It makes the text much smaller compared to the text present in included graphics, which is stretched across the whole size of a page.

The proposed margins in conjunction with the A5 format work quire well. There are still around 10 code snippets across the whole PDF which happened to be line-wrapped, though.

The original title for the generated book was "Table of Contents". :-)

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,0 +1,14 @@
+{
+    "title": "Linux Insides",
+    "author" : "0xAX",
+    "pdf": {
+        "paperSize": "a5",
+        "margin":
+        {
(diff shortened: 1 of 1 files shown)


0xAX:
@kyselejsyrecek thank you for PR!

PR #877, opened 2025-09-04
Work around broken export of SVGs not compliant with the 1.1 standard

kyselejsyrecek:
**Description**

SVGs generated with draw.io contain XHTML tags and are thus not compliant with the SVG 1.1 standard. While rendered correctly by web browsers and Inkscape, the svgexport package used by the gitbook executable to export SVG images fails to do so as ImageMagick and rsvg-convert do. This would result in black rectangles covering most of SVG images present in the Booting chapter.

To prevent this, let's just use Inkscape CLI to convert source SVG images directly to PNGs before building the book with gitbook. Unfortunately, it seems that inkscape has to be invoked separately for each SVG, as the interactive mode toggled with the --shell switch seems to be too buggy to do that in one run in this outdated version of Inkscape.

Optionally, the image resolution may be increased to 300 DPI if necesary. That would result in slightly larger e-book files.

Further changes:

- Use Bash as a more user-friendly default shell for the Docker container.
- The Docker container was updated. Installed packages were updated and packages bzip2, calibre and inkscape required for gitbook export were installed.

**Related issues**

- https://github.com/0xAX/linux-insides/issues/865
- 
image


kyselejsyrecek:
Oh, so apparently the workflow _does_ run even for pull requests from forks. :-)

0xAX:
Looked at the generated PDF, looks great!

PR #915, opened 2026-03-29
init-1: rework the first part about the kernel initialization

0xAX:
**Description**

This commit updates the first part related to the Linux kernel initialization process to adjust it to the modern Linux kernel versions.


Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
-Kernel initialization. Part 1.
-================================================================================
+# Linux kernel initialization - Part 1
 
-First steps in the kernel code
---------------------------------------------------------------------------------
+The previous [post](https://github.com/0xAX/linux-insides/blob/master/Booting/linux-bootstrap-6.md) was the final part of the chapter describing the [Linux kernel boot process](https://github.com/0xAX/linux-insides/tree/master/Booting). In that chapter, we looked at the kernel boot process path step by step, starting from the very first actions and instructions executed after the system powers on, through the bootloader, and finally into the Linux kernel setup code.
```
klaudiagrz:
```suggestion
The previous [post](https://github.com/0xAX/linux-insides/blob/master/Booting/linux-bootstrap-6.md) was the final part of the chapter describing the [Linux kernel boot process](https://github.com/0xAX/linux-insides/tree/master/Booting). In that chapter, we looked at the kernel boot process step by step, starting from the very first actions and instructions executed after the system powers on, through the bootloader, and finally into the Linux kernel setup code.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+All upcoming parts of this chapter will be dedicated to exploring how the kernel initializes itself before launching the very first userspace process with [PID 1](https://en.wikipedia.org/wiki/Process_identifier). It can be hard to believe, but even after all the code we have already seen, there is still a lot to do. The kernel must initialize all core subsystems, finish setting up memory management, detect hardware, load drivers for it, and perform many other tasks before userspace code will be able to run.
 
-```assembly
-jmp	*%rax
-```
+Our journey begins at the kernel entry point located in [arch/x86/kernel/head_64.S](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/head_64.S). From there we will move step by step through the early initialization code, looking at the first actions performed by the kernel, including:
```
klaudiagrz:
```suggestion
Our journey begins at the kernel entry point located in [arch/x86/kernel/head_64.S](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/head_64.S). From there, we will move step by step through the early initialization code, looking at the first actions performed by the kernel, including:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-The ADDR keyword above returns the absolute address (here means virtual address) of the named section. The AT keyword above specifies the load address (here means physical address) of the section. The full syntax of section definition is defined in the [Using ld The GNU linker](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_21.html). 
+<!-- https://github.com/torvalds/linux/raw/refs/heads/master/arch/x86/kernel/head_64.S#L36-L38 -->
+```assembly
+	__INIT
+	.code64
+SYM_CODE_START_NOALIGN(startup_64)
+```
 
+It is placed in the `.init.text` section by the `__INIT` macro. This section will exists only during early initialization and as soon as the kernel will finish the initialization process, it will be freed.
```
klaudiagrz:
```suggestion
It is placed in the `.init.text` section by the `__INIT` macro. This section will exist only during early initialization, and as soon as the kernel finishes the initialization process, it will be freed.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-```C
-	load_delta = physaddr - (unsigned long)(_text - __START_KERNEL_map);
+<!-- https://github.com/torvalds/linux/raw/refs/heads/master/arch/x86/kernel/head_64.S#L62-L62 -->
+```assembly
+	leaq	__top_init_kernel_stack(%rip), %rsp
 ```
 
-As a result, `load_delta` contains the delta between the address compiled to run at and the address actually loaded.
+It points not just at a random address, but belongs to the init task, the very first kernel thread (PID 0), which will eventually become the idle process. The stack is `16384` bytes size and its top is defined in the kernel image by the [linker script](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/vmlinux.lds.S). The `__top_init_kernel_stack` symbol is defined as:
```
klaudiagrz:
```suggestion
It points not just at a random address, but belongs to the init task, the very first kernel thread (PID 0), which will eventually become the idle process. The stack is `16384` bytes in size, and its top is defined in the kernel image by the [linker script](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/vmlinux.lds.S). The `__top_init_kernel_stack` symbol is defined as:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
-```C
-	if (load_delta & ~PMD_PAGE_MASK)
-		for (;;);
+<!-- https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/arch/x86/kernel/vmlinux.lds.S#L183-L183 -->
+```linker-script
+		__top_init_kernel_stack = __end_init_stack - TOP_OF_KERNEL_STACK_PADDING - PTREGS_SIZE;
 ```
 
-If `_text` address is not aligned for `2` megabytes, we enter infinite loop. The `PMD_PAGE_MASK` indicates the mask for `Page middle directory` (read [Paging](https://0xax.gitbook.io/linux-insides/summary/theory/linux-theory-1) about it) and is defined as:
+The two subtractions reserve space at the very top of the stack:
```
klaudiagrz:
```suggestion
Two subtractions reserve space at the very top of the stack:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-```C
-#define PMD_PAGE_MASK           (~(PMD_PAGE_SIZE-1))
-```
+- `TOP_OF_KERNEL_STACK_PADDING` - currently reserved for [Intel FRED](https://www.intel.com/content/www/us/en/content-details/779982/flexible-return-and-event-delivery-fred-specification.html)
+- `PTREGS_SIZE` - leaves room for the register frame that gets pushed during interrupt handling
 
-where `PMD_PAGE_SIZE` macro is defined as:
+With the stack ready, the kernel can now call functions and handle interrupts. But there is one more thing to clean up before moving on. The `gs` register is used by the kernel to access [per-CPU data structures](../Concepts/linux-cpu-1.md). Right now, it may still contain whatever value the decompressor left. If that garbage value were used as a base for a per-CPU access, the kernel would read from or write to some wrong memory location. So the kernel zeroes it out:
```
klaudiagrz:
```suggestion
With the stack ready, the kernel can now call functions and handle interrupts. But there is one more thing to clean up before moving on. The kernel uses the `gs` register to access [per-CPU data structures](../Concepts/linux-cpu-1.md). Right now, it may still contain whatever value the decompressor left. If that garbage value were used as a base for a per-CPU access, the kernel would read from or write to the wrong memory location. So the kernel zeroes it out:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
-NEXT_PAGE(early_top_pgt)
-	.fill	512,8,0
-	.fill	PTI_USER_PGD_FILL,8,0
+The main reason to load these descriptors is how they are used in long mode. Although the base and limit values of the code segment are ignored and no longer used for address calculations, the other fields still function normally. For example, privilege checks are still performed.
 
-NEXT_PAGE(level3_kernel_pgt)
-	.fill	L3_START_KERNEL,8,0
-	.quad	level2_kernel_pgt - __START_KERNEL_map + _KERNPG_TABLE_NOENC
-	.quad	level2_fixmap_pgt - __START_KERNEL_map + _PAGE_TABLE_NOENC
+The new Global Descriptor Table is loaded by the following call:
```
klaudiagrz:
```suggestion
The following call loads the new Global Descriptor Table:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+<!-- https://github.com/torvalds/linux/raw/refs/heads/master/arch/x86/kernel/head_64.S#L62-L62 -->
+```assembly
+	call	__pi_startup_64_setup_gdt_idt
+```
 
-NEXT_PAGE(level2_fixmap_pgt)
-	.fill	506,8,0
-	.quad	level1_fixmap_pgt - __START_KERNEL_map + _PAGE_TABLE_NOENC
-	.fill	5,8,0
+Before we look at this function, there is a interesting detail about how it is called. You will not find its definition if you try to grep the Linux kernel source code:
```
klaudiagrz:
```suggestion
Before we look at this function, there is an interesting detail about how it is called. You will not find its definition if you try to grep the Linux kernel source code:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+# code, or code that has explicitly been made accessible to it via a symbol
+# alias.
+#
+$(obj)/%.pi.o: OBJCOPYFLAGS := --prefix-symbols=__pi_
+$(obj)/%.pi.o: $(obj)/%.o FORCE
+	$(call if_changed,objcopy)
+```
 
-You can read more about it in the [Paging](https://0xax.gitbook.io/linux-insides/summary/theory/linux-theory-1) part.
+This is done to prevent the compiler from generating any absolute address references. As you can remember from the last parts of the previous chapter, the kernel uses identity-mapping page tables at this point. These page tables map 1:1 the first four gigabytes of memory. On the other hand, the kernel is linked to the high memory virtual addresses starting from the `0xffffffff80000000` address. In this case, any absolute address reference would cause a [page fault](https://en.wikipedia.org/wiki/Page_fault) which the kernel is not ready to handle yet.
```
klaudiagrz:
```suggestion
This is done to prevent the compiler from generating any absolute address references. As you can remember from the last parts of the previous chapter, the kernel uses identity-mapping page tables at this point. These page tables map 1:1 the first four gigabytes of memory. On the other hand, the kernel is linked to the high memory virtual addresses starting from the `0xffffffff80000000` address. In this case, any absolute address reference would cause a [page fault](https://en.wikipedia.org/wiki/Page_fault), which the kernel is not ready to handle yet.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-Now, after we saw the definitions of these symbols, let's get back to the code. Next we initialize last entry of `pgd` with `level3_kernel_pgt`:
+Now that we know where the new GDT is loaded, let's take a look at the `startup_64_setup_gdt_idt` function. It does the three following things:
```
klaudiagrz:
```suggestion
Now that we know where the new GDT is loaded, let's take a look at the `startup_64_setup_gdt_idt` function. It does the following things:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+This code is more comfortable to read from the end. The `lretq` instruction pops two values from the stack:
 
-Next we initialize `pgtable_flags` which will be used when initializing `p*d` entries later:
+- The first into `rip` register
+- The second into `cs` register
 
-```C
-	pgtable_flags = _KERNPG_TABLE_NOENC + sme_get_me_mask();
-```
+After execution, the CPU continues at `.Lon_kernel_cs` symbol which is located right after the `lretq` instruction, ensuring the kernel is running with the correct code segment from its own GDT.
```
klaudiagrz:
```suggestion
After execution, the CPU continues at `.Lon_kernel_cs` symbol, which is located right after the `lretq` instruction, ensuring the kernel is running with the correct code segment from its own GDT.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
-Next we fill two entries of `pgd` with `pud` plus `pgtable_flags` which we initialized above:
+> [!NOTE]
+> The code in this section also handles [5-level paging](https://en.wikipedia.org/wiki/Intel_5-level_paging) and [AMD Secure Memory Encryption](https://developer.amd.com/sev/). We will not cover these topics here to keep things focused on the core page table setup.
 
-```C
-	i = (physaddr >> PGDIR_SHIFT) % PTRS_PER_PGD;
-	pgd[i + 0] = (pgdval_t)pud + pgtable_flags;
-	pgd[i + 1] = (pgdval_t)pud + pgtable_flags;
-```
+The next big step after the GDT setup is to adjust the page tables. This is one of the most critical parts of the early initialization. The kernel needs to build page tables that map it at its high virtual address, create a temporary identity mapping so that switching to the new page tables does not cause a page fault, and then perform the switch.
```
klaudiagrz:
```suggestion
The next big step after the GDT setup is to adjust the page tables. This is one of the most critical parts of the early initialization, where the kernel needs to:

- Build page tables that map it at its high virtual address
- Create a temporary identity mapping so that switching to the new page tables does not cause a page fault
- Perform the switch
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-```C
-#define PGDIR_SHIFT     39
-#define PTRS_PER_PGD	512
-#define PUD_SHIFT       30
-#define PTRS_PER_PUD	512
-#define PMD_SHIFT       21
-#define PTRS_PER_PMD	512
-```
+Before we dive in, a quick reminder about the Linux kernel x86_64 page table hierarchy. To translate a virtual address to the corresponding physical address, the CPU walks through up to four levels of page tables:
```
klaudiagrz:
```suggestion
Before we dive in, a quick reminder about the Linux kernel x86_64 page table hierarchy. To translate a virtual address to the corresponding physical address, the CPU goes through up to four levels of page tables:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+<!-- https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/include/asm-generic/pgtable-nop4d.h#L9-L9 -->
 ```C
-	i = (physaddr >> PUD_SHIFT) % PTRS_PER_PUD;
-	pud[i + 0] = (pudval_t)pmd + pgtable_flags;
-	pud[i + 1] = (pudval_t)pmd + pgtable_flags;
+typedef struct { pgd_t pgd; } p4d_t;
 ```
 
-Next we initialize `pmd_entry` and filter out unsupported `__PAGE_KERNEL_*` bits:
+The kernel can also use 2 MB or so-called large pages that stop at the PMD level, skipping the PTE level entirely. This is what the early kernel page tables use. In the code below, you will see how the kernel builds these page tables:
```
klaudiagrz: I expected to see some code there, not bullet points:

Image

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+But this is not where the kernel is supposed to be mapped. On x86_64, the virtual address space is huge - `2^47` bytes with 4-level paging, and it is split into two halves:
 
-```C
-	pmd = fixup_pointer(level2_kernel_pgt, physaddr);
-	for (i = 0; i < PTRS_PER_PMD; i++) {
-		if (pmd[i] & _PAGE_PRESENT)
-			pmd[i] += load_delta;
-	}
-```
+- The lower half which covers addresses starting from `0x0` belongs to user-space processes
```
klaudiagrz:
```suggestion
- The lower half, which covers addresses starting from `0x0`, belongs to user-space processes
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+- The upper half is the kernel's address space
 
-Next we remove the memory encryption mask to obtain the true physical address (remember that `load_delta` includes the mask):
+This region where kernel is mapped is usually called `high memory`. This is where the linker places all kernel symbols when `vmlinux` is built. We can verify this with `readelf`:
```
klaudiagrz:
```suggestion
This region where the kernel is mapped is usually called `high memory`. This is where the linker places all kernel symbols when `vmlinux` is built. We can verify this with `readelf`:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-```C
-	*fixup_long(&phys_base, physaddr) += load_delta - sme_get_me_mask();
+```bash
+readelf -S vmlinux | grep "  \[ 1\] \.text"
+  [ 1] .text             PROGBITS         ffffffff81000000  00200000
 ```
 
-`phys_base` must match the first entry in `level2_kernel_pgt`.
+The whole virtual memory map of the Linux kernel for x86_64 can be found in the [Documentation/arch/x86/x86_64/mm.rst](https://github.com/torvalds/linux/blob/master/Documentation/arch/x86/x86_64/mm.rst) document. I know. This table can look intimidating at first glance but do not worry. We will refer back to specific regions of memory as they become relevant. For now, we are interested only in the last few entries at the very bottom of the table which describes the kernel code mapping areas:
```
klaudiagrz:
```suggestion
You can find the whole virtual memory map of the Linux kernel for x86_64 in the [Documentation/arch/x86/x86_64/mm.rst](https://github.com/torvalds/linux/blob/master/Documentation/arch/x86/x86_64/mm.rst) document. I know. This table can look intimidating at first glance, but do not worry. We will refer back to specific regions of memory as they become relevant. For now, we are interested only in the last few entries at the very bottom of the table, which describe the kernel code mapping areas:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
 
-As final step of `__startup_64` function, we encrypt the kernel (if SME is active) and return the SME encryption mask to be used as a modifier for the initial page directory entry programmed into `cr3` register:
+| Start addr       | Offset    | End addr         | Size    | Description                                                                                                                                                                                                              |
```
klaudiagrz:
```suggestion
| Start address       | Offset    | End address         | Size    | Description                                                                                                                                                                                                              |
```

klaudiagrz: Is this table a copy-paste from somewhere?

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
+| ffffffffff600000 | -10 MB    | ffffffffff600fff | 4 kB    | legacy vsyscall ABI                                                                                                                                                                                                      |
+| ffffffffffe00000 | -2 MB     | ffffffffffffffff | 2 MB    | ... unused hole                                                                                                                                                                                                          |
 
-```C
-	sme_encrypt_kernel(bp);
-	return sme_get_me_mask();
-```
+Every function, every global variable in the kernel was linked to run at these high addresses. But our current page tables know nothing about this memory region since they cover only the low memory range. If the kernel tried to jump to its linked address right now, the CPU would trigger a page fault because there is no page table entry for it.
```
klaudiagrz:
```suggestion
Every function and every global variable in the kernel was linked to run at these high addresses. But our current page tables know nothing about this memory region since they cover only the low memory range. If the kernel tried to jump to its linked address right now, the CPU would trigger a page fault because there is no page table entry for it.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,639 @@
...
-	addq	$(early_top_pgt - __START_KERNEL_map), %rax
-	jmp 1f
-```
+The current page tables from the decompressor look like this:
 
-which adds physical address of `early_top_pgt` to `rax` register so that `rax` register contains sum of the address and the SME encryption mask.
+![early-page-tables](./images/early-page-tables.svg)
 
-That's all for now. Our early paging is prepared and we just need to finish last preparation before we will jump into kernel entry point.
+After this, the new page tables will look like this:
```
klaudiagrz: After what?

klaudiagrz (commented):
Checked till "Building the page tables"

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,613 @@
...
-
-The `init_thread_union` is represented by the `thread_union`. And the `thread_union` is defined in the [include/linux/sched.h](https://github.com/torvalds/linux/blob/master/include/linux/sched.h) file like the following:
+This is exactly the technique used at the beginning of the `__startup_64` function to get the physical address corresponding to the base of the kernel's virtual address space denoted by the `phys_base` and the difference between where the kernel was linked to run and where it was actually loaded denoted by the `load_delta`:
```
klaudiagrz: This sentence is very long; consider splitting it

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
+SYM_DATA_END(early_top_pgt)
+SYM_PIC_ALIAS(early_top_pgt)
 ```
-.data : AT(ADDR(.data) - LOAD_OFFSET) {
-	...
-	INIT_TASK_DATA(THREAD_SIZE)
-	...
-} :data
+
+The first `511` entries of the top level page table are empty and only the last one points to `level3_kernel_pgt`. At first glance you may think that the single entry is not enough, but this perfectly makes sense. Each PGD entry covers 512 GB of memory, and the kernel is mapped in the very last 2 GB of the virtual address space. Entry `511` is exactly the one that covers that memory region. The remaining entries are empty for now, but will be filled in the `__startup_64` function with the identity mapping pages needed for the `cr3` register switch. We will see that a little later, but first comes the adjusting of the top-level page entry:
```
klaudiagrz:
```suggestion
The first `511` entries of the top level page table are empty, and only the last one points to `level3_kernel_pgt`. At first glance, you may think that the single entry is not enough, but this perfectly makes sense. Each PGD entry covers 512 GB of memory, and the kernel is mapped in the very last 2 GB of the virtual address space. Entry `511` is exactly the one that covers that memory region. The remaining entries are empty for now, but will be filled in the `__startup_64` function with the identity mapping pages needed for the `cr3` register switch. We will see that a little later, but first comes the adjusting of the top-level page entry:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
-```assembly
-lgdt	early_gdt_descr(%rip)
+Or just:
+
+```
+level3_kernel_pgt + p2v_offset + _PAGE_TABLE_NOENC
 ```
 
-where the `early_gdt_descr` is defined as:
+Which is the physical address of the `level3_kernel_pgt` and its flags. This is exactly what must be placed in a page table entry.
```
klaudiagrz:
```suggestion
This is the physical address of the `level3_kernel_pgt` and its flags, and that's exactly what must be placed in a page table entry.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
+<!-- https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/arch/x86/boot/startup/map_kernel.c#L139-L140 -->
+```C
+	for (i = FIXMAP_PMD_TOP; i > FIXMAP_PMD_TOP - FIXMAP_PMD_NUM; i--)
+		level2_fixmap_pgt[i].pmd += load_delta;
+```
 
-As we loaded new Global Descriptor Table, we reload segments as we did it every time:
+At this point, the high-memory kernel mapping is complete. But if the kernel would try to switch to these page tables right now, things would immediately go wrong. The CPU is still executing at a low physical address while the kernel just built page tables that do not map it. The very next instruction fetch after updating the page tables would trigger a [page fault](https://en.wikipedia.org/wiki/Page_fault), and the kernel has no handler ready to catch it.
```
klaudiagrz:
```suggestion
At this point, the high-memory kernel mapping is complete. But if the kernel would try to switch to these page tables right now, things would immediately go wrong. The CPU is still running at a low physical address while the kernel has just built page tables that do not map it. The very next instruction fetch after updating the page tables would trigger a [page fault](https://en.wikipedia.org/wiki/Page_fault), and the kernel has no handler ready to catch it.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
 
-```assembly
-	xorl %eax,%eax
-	movl %eax,%ds
-	movl %eax,%ss
-	movl %eax,%es
-	movl %eax,%fs
-	movl %eax,%gs
-```
+To avoid this problem, the kernel needs to build a temporary identity mapping page tables. By also mapping the low physical address range where the CPU is currently executing, the kernel makes sure that both the old address (where execution is right now) and the new high address (where the kernel should be mapped) are valid at the moment of the switch. Once the kernel jumps to its high virtual address, this identity mapping will have served its purpose and can be discarded.
```
klaudiagrz:
```suggestion
To avoid this problem, the kernel needs to build a temporary identity mapping page tables. By also mapping the low physical address range where the CPU is currently running, the kernel makes sure that both the old address (where execution is right now) and the new high address (where the kernel should be mapped) are valid at the moment of the switch. Once the kernel jumps to its high virtual address, this identity mapping will have served its purpose and can be discarded.
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
 
-We need to see last preparations before we can see "kernel entry point" - start_kernel function from the [init/main.c](https://github.com/torvalds/linux/blob/master/init/main.c).
+After switching to the new page table, the old [TLB](https://en.wikipedia.org/wiki/Translation_lookaside_buffer) entries are stale now. If a stale entry were used, the CPU would silently translate an address to the wrong physical location. So the kernel needs to flush the TLB.
 
-First of all we can see some checks in the `x86_64_start_kernel` function:
+The standard way to do this on x86_64 is to toggle the `PGE` (Page Global Enable) bit in the `cr4` [control register](https://en.wikipedia.org/wiki/Control_register). Clearing this bit invalidates all global TLB entries, and setting it back re-enables global translations with fresh entries from the new page tables. But `cr4` contains many other important bits, so the kernel cannot just zero the whole register. It first builds a mask of the bits that must survive the flush:
```
klaudiagrz: How is the list related to the previous paragraph? Are these the bits you mention in the last sentence? In this case, I would add something like "These bits are:"  and then list.

Image

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
 
-In the next part we will see initialization of the early interruption handlers, kernel space memory mapping and a lot more.
+In the next part, we will continue to dive into this interesting process and will see the continuation of the x86_64 architecture-specific initialization before the kernel jumps to the "generic" entry point - `start_kernel`.
 
-**Please note that English is not my first language and I am really sorry for any inconvenience. If you found any mistakes please send me PR to [linux-insides](https://github.com/0xAX/linux-insides).**
+## Links
 
-Links
---------------------------------------------------------------------------------
+Here is the list of the links that you may find useful during reading of this chapter:
```
klaudiagrz:
```suggestion
Here is the list of the links that you may find useful when reading this chapter:
```

Initialization/linux-initialization-1.md
```diff
@@ -1,692 +1,623 @@
...
-* [NX](http://en.wikipedia.org/wiki/NX_bit)
-* [ASLR](http://en.wikipedia.org/wiki/Address_space_layout_randomization)
+- [Linux kernel boot process](https://github.com/0xAX/linux-insides/tree/master/Booting)
+- [Real mode](https://en.wikipedia.org/wiki/Real_mode)
+- [Protected mode](https://en.wikipedia.org/wiki/Protected_mode)
+- [Long mode](https://en.wikipedia.org/wiki/Long_mode)
+- [GNU linker documentation](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_21.html)
+- [Paging](https://en.wikipedia.org/wiki/Memory_paging)
+- [model-specific register](https://en.wikipedia.org/wiki/Model-specific_register)
+- [control register](https://en.wikipedia.org/wiki/Control_register)
```
klaudiagrz:
```suggestion
- [Model-specific register](https://en.wikipedia.org/wiki/Model-specific_register)
- [Control register](https://en.wikipedia.org/wiki/Control_register)
```

PR #954, opened 2026-09-09
general: update the list of users support linux-insides

0xAX:
**Description**

This PR updates the list of supporters of linux-insides on Patreon.


klaudiagrz (approved):
great idea!

